PR #101, opened 2019-02-15
how-to-sanitize-user-generated-content-in-ftw

Gnito:
~~I haven't yet reordered guides, but my idea was to add this just before CSP guide.~~


src/docs/guides/11-how-to-sanitize-user-generated-content-in-ftw/index.md
```diff
@@ -0,0 +1,71 @@
...
+
+We have added an example of how user-generated content could be
+sanitized when entities are queried and updated to Redux store. Those
+examples can be found from the utility file
+[src/util/sanitize.js](https://github.com/sharetribe/flex-template-web/blob/master/src/util/sanitize.js).
+
+You should modify those sanitize functions (e.g. `sanitizeUser`) to
+include any extendedData you have created for your marketplace. This is
+important if the data is used as a props/attribute in components (e.g.
+`<div attr=...ESCAPE UNTRUSTED DATA BEFORE PUTTING HERE...>content`). As
```
kpuputti: I guess this depends on the attribute, right? So something like `<a data-user-href={user.publicData.link}>` is fine but `<a href={user.publicData.link}>` is not.

So React ensures that you cannot do something like `user.publicData.link='#" onClick="javascript:alert(\'oh no\')" hacked="true'` which would end up as `<a data-user-href="#" onClick="javascript:alert(\'oh no\')" hacked="true">` in a simple template language that only does string concatenation and not actual DOM operations.

Am I correct? This is mostly about understanding the issue, I don't think we even need to mention that some attributes are safe...

Gnito: Yes. 
Your example would convert to something like
```
<a data-user-href="#" onClick="javascript:alert('oh no')" hacked="true">
```

PR #113, opened 2019-03-04
Add a note about GA not working with hot loading.

Gnito:


src/docs/guides/14-how-to-set-up-analytics-for-ftw/index.md
```diff
@@ -18,6 +18,11 @@ tracker ID to the environment variable `REACT_APP_GOOGLE_ANALYTICS_ID`;
 
 That's it! You're good to go.
 
+> Note: Google Analytics doesn't work in hot-loading environment!
+>       <br/>
```
kpuputti: Did this HTML element come from the GitHub editor? It is rendered as HTML, but looks a bit weird in the Markdown source. Maybe just use an extra empty line there?

kpuputti:
```suggestion
>
```

Gnito:
@kpuputti I wasn't aware of other ways to add line-breaks to blocks, but 2 spaces at the end of a row does the same!

kpuputti:
@Gnito Let's not use trailing whitespace, that is very unexpected and those might be removed accidentally.

Gnito:
@kpuputti come on - it's either or. I'm not going to merge something that doesn't have a line-break. The first sentence (the actual warning) disappears to the noise of _"tl:dr"_ if it's not separated from the explanation.

PR #130, opened 2019-03-28
typography and layout CSS finetuning

jannekoivistoinen:
@kpuputti simple CSS changes to the layout to make it bit easier to read.

- Biggest change is going from 3-column layout to 1-column layout in ArticleIndexPage, imo it's easier to read vertically than scan in Z-format with this many articles. 

Future improvements could be categorising the Articles within the page for easier skimming and adding a universal predictive search.


src/components/ArticlePage/Toc.js
```diff
@@ -11,10 +11,9 @@ const TocLi = styled(Li).attrs({
   // Side navigation hidden on small viewport
 
   @media (min-width: ${baselineBreakpoint}px) {
-    line-height: 32px;
-
-    // Offset baseline
-    top: 2px;
+    line-height: 24px;
+    padding-bottom: ${baselineLarge}px;
+    font-size: 15px;
```
kpuputti: This is now out of baseline alignment. Add the offset:

```suggestion
    font-size: 15px;
    
    // Offset baseline
    top: -1px;
```

src/components/Box.js
```diff
@@ -11,13 +11,19 @@ const Box = styled.section`
...
   box-shadow: 0 8px 16px 0 rgba(0,0,0,0.05);
+  transition: all 0.05s ease-out;
 
   @media (min-width: ${baselineBreakpoint}px) {
     padding-top: ${4 * baselineLarge}px;
     padding-bottom: ${5 * baselineLarge}px;
-    padding-left: ${props => props.theme.contentPaddingLarge}px;
-    padding-right: ${props => props.theme.contentPaddingLarge}px;
+    padding-left: ${5 * baselineLarge}px;
+    padding-right: ${5 * baselineLarge}px;
```
kpuputti: Again, baseline variables shouldn't be used for horizontal spacing. If the content padding theme values are not good, you can just inline the values here.

src/components/Box.js
```diff
@@ -11,13 +11,19 @@ const Box = styled.section`
...
   }
+
+  :hover {
+		box-shadow: 0 16px 32px 0 rgba(0,0,0,0.1);
+		transform: scale(1.01);
+	}
```
kpuputti: Indentation formatting still off.

src/components/GlobalStyle.js
```diff
@@ -4,6 +4,7 @@ const GlobalStyle = createGlobalStyle`
   body {
     background-color: ${props => props.theme.backgroundColor};
     overflow-y: scroll;
+    -webkit-font-smoothing: antialiased;
```
kpuputti: What is this doing and is it necessary? I've heard some arguments why this shouldn't be used, but I don't have up to date info on what the current best practices are.

This might make fonts look better on MacOS etc., but is it harming other browsers?

src/components/LandingPage/LandingPage.js
```diff
@@ -28,11 +28,9 @@ export const IntroHeading = styled(H1)`
...
 
   @media (min-width: ${baselineBreakpoint}px) {
-    margin-top: ${14 * baselineLarge}px;
-    margin-left: ${props =>
-      props.theme.contentPaddingLarge + grid.sideMargin}px;
-    margin-right: ${props =>
-      props.theme.contentPaddingLarge + grid.sideMargin}px;
+    margin-top: ${10 * baselineLarge}px;
+    margin-left: ${5 * baselineLarge + grid.sideMargin}px;
+    margin-right: ${5 * baselineLarge + grid.sideMargin}px;
```
kpuputti: Same here, no baseline vars for horizontal spacing. That's why we have the content padding vars in the theme.

src/components/LandingPage/LandingPage.js
```diff
@@ -43,10 +41,8 @@ export const IntroDescription = styled(P)`
 
   @media (min-width: ${baselineBreakpoint}px) {
     margin-top: ${2 * baselineLarge}px;
-    margin-left: ${props =>
-      props.theme.contentPaddingLarge + grid.sideMargin}px;
-    margin-right: ${props =>
-      props.theme.contentPaddingLarge + grid.sideMargin}px;
+    margin-left: ${5 * baselineLarge + grid.sideMargin}px;
+    margin-right: ${5 * baselineLarge + grid.sideMargin}px;
```
kpuputti: And here ^

src/components/LandingPage/LandingPage.js
```diff
@@ -91,10 +87,8 @@ const Paragraph = styled(P).attrs({
   @media (min-width: ${baselineBreakpoint}px) {
     margin-top: ${7 * baselineLarge}px;
     margin-bottom: ${13 * baselineLarge}px;
-    margin-left: ${props =>
-      props.theme.contentPaddingLarge + grid.sideMargin}px;
-    margin-right: ${props =>
-      props.theme.contentPaddingLarge + grid.sideMargin}px;
+    margin-left: ${5 * baselineLarge + grid.sideMargin}px;
+    margin-right: ${5 * baselineLarge + grid.sideMargin}px;
```
kpuputti: Same here ^

src/components/LandingPage/LandingPage.js
```diff
@@ -112,10 +106,8 @@ const CollectionsHeading = styled(H6)`
 
   @media (min-width: ${baselineBreakpoint}px) {
     margin-top: ${12 * baselineLarge}px;
-    margin-left: ${props =>
-      props.theme.contentPaddingLarge + grid.sideMargin}px;
-    margin-right: ${props =>
-      props.theme.contentPaddingLarge + grid.sideMargin}px;
+    margin-left: ${5 * baselineLarge + grid.sideMargin}px;
+    margin-right: ${5 * baselineLarge + grid.sideMargin}px;
```
kpuputti: Same here ^

src/components/LandingPage/LandingPage.js
```diff
@@ -142,10 +134,8 @@ const CategoriesHeading = styled(H6)`
 
   @media (min-width: ${baselineBreakpoint}px) {
     margin-top: ${10 * baselineLarge}px;
-    margin-left: ${props =>
-      props.theme.contentPaddingLarge + grid.sideMargin}px;
-    margin-right: ${props =>
-      props.theme.contentPaddingLarge + grid.sideMargin}px;
+    margin-left: ${5 * baselineLarge + grid.sideMargin}px;
+    margin-right: ${5 * baselineLarge + grid.sideMargin}px;
```
kpuputti: And here ^

src/components/GlobalStyle.js
jannekoivistoinen: @kpuputti interesting! First time I hear about this and we have been using this everywhere, FTW, Console, marketing-site etc. This makes the fonts look crispier and thus increases the readability because of the crispier edge of the pixels. 

I'll suggest that we put it in now, and if it causing issues in Windows or Linux, then we can always target only MacOs individually with this.

src/components/Box.js
```diff
@@ -11,13 +11,19 @@ const Box = styled.section`
...
   box-shadow: 0 8px 16px 0 rgba(0,0,0,0.05);
+  transition: all 0.05s ease-out;
 
   @media (min-width: ${baselineBreakpoint}px) {
-    padding-top: ${4 * baselineLarge}px;
-    padding-bottom: ${5 * baselineLarge}px;
-    padding-left: ${props => props.theme.contentPaddingLarge}px;
-    padding-right: ${props => props.theme.contentPaddingLarge}px;
+    padding-top: 32px;
+    padding-bottom: 40px;
```
kpuputti: Top/bottom should still use the baseline vars.

src/components/GlobalStyle.js
kpuputti: I remember researching it a bit earlier as well, and maybe even discussing with @Gnito . Not that I'm not only referring to all browsers on MacOS, different browsers might behave differently also on MacOS.

The usual accessibility argument is that browsers should know by default what's the best way to render stuff. But if this has clear visible benefits, let's just keep it for now.

src/components/Box.js
kpuputti: Could you maybe fix this manually if the `yarn run format` doesn't work here?

src/components/GlobalStyle.js
jannekoivistoinen: @kpuputti visually I think it has a clear benefit. In my opinion, only Safari can render the fonts correctly and all other browsers aren't so great out of the box.

PR #133, opened 2019-04-02
Content fixes

jannekoivistoinen:
Increasing readability and being consistent:
- Add a HR divider before H2
- If the in-page title was H1, that has been changed to H2
- Add links to the "Getting started" article's titles
- Minor other improvements for consistency


src/docs/background/04-getting-started/index.md
```diff
@@ -11,27 +11,33 @@ ingress:
 published: true
 ---
 
-## Introducing Flex
+## [Introducing Flex](/background/introducing-flex/)
```
kpuputti: We originally had these heading links, but decided to drop them and added the "Read more" links. I think the main reason was that people were't aware that the headings were links, and only having those as links wasn't enough.

I guess there's no harm in having both, but do you @thomasmalbaux if there was some other reason why we didn't want the heading links? I remember you took part in the discussion.

src/docs/background/04-getting-started/index.md
```diff
@@ -11,27 +11,33 @@ ingress:
...
 
 Introduction to the three main components of Flex: your marketplace user
 interface, the Marketplace API, and Console. We explain the role each of
 these components plays and how they help you to build a unique
 marketplace platform. [Read more »](/background/introducing-flex/)
 
-## How to know if Flex is the right fit for your marketplace idea
+---
```
kpuputti: I don't think it makes sense to require all article authors to add these. This will only make the content inconsistent as people forget adding these lines. Could we just style the article `h2` elements to have a top border?

src/docs/guides/08-how-to-use-google-maps-in-ftw/index.md
```diff
@@ -70,12 +76,14 @@ current location from
 Search for variables: `suggestCurrentLocation` and
 `currentLocationBoundsDistance`.
 
-## 4. Change components: use Google Map versions instead of Mapbox
+---
+
+## 5. Change components: use Google Map versions instead of Mapbox
```
kpuputti: Be very careful when changing headings in the site. The anchor links are automatically generated from the headings, so changing a heading also breaks existing links within the site or in support tickets. I see that the numbering was indeed wrong and it probably makes sense to fix it, but just FYI.

src/docs/background/04-getting-started/index.md
jannekoivistoinen: @kpuputti hmm, yeah maybe having a border-top for H2 (excl. the first H2 on the page) would make it easier for the content writers - I'll create the classes and remove the lines! 🍬

src/docs/background/04-getting-started/index.md
jannekoivistoinen: I think they should be links, as then we stay consistent with the rest of the site - in the homepage titles are links, in the category pages the titles are links and should it be in the collection pages for consistency.

Personally I always try click the title as it has a bigger hit area than the "read more".

src/docs/guides/08-how-to-use-google-maps-in-ftw/index.md
jannekoivistoinen: @kpuputti thanks, very good to know! So 'yarn run format' doesn't re-generate the anchor links? How do they get generated and how to fix them if anchor links are broken?

kpuputti: Nothing will automatically fix the links. The anchor is automatically generated from the heading, so if the headings changes, the anchor will change accordingly. There is a plugin that does this automatically.

Internal links are the "easy" part, and there isn't too many of those. I've used https://validator.w3.org/checklink every once in a while to crawl the site for broken links and anchors. But I assume the main problem is links that are not in the site itself, but e.g. in support and other sites.

src/docs/background/04-getting-started/index.md
thomasmalbaux: I'm not against having them as links however without the Read more, I think there is no way to know that they are links and describe and article. They just look title headers. So again as long as there is some clear sign (read more, something), I'm happy.

PR #138, opened 2019-04-08
replacing content-fixes branch

jannekoivistoinen:


src/components/ArticlePage/MarkdownHtml.js
```diff
@@ -89,8 +89,15 @@ const Html = styled.div`
   }
   h2 {
     ${H4.styles}
-    margin: ${2 * baselineSpacing}px 0 ${baselineSpacing}px 0;
+    margin: ${1.5 * baselineSpacing}px 0 ${baselineSpacing}px 0;
```
kpuputti: You should rather use a multiple of the `baselineSmall` variable instead of calculating something from the `baselineSpacing`, that is just a quick shorthand for cases where I wasn't sure what to put there.

```suggestion
    margin: ${6 * baselineSmall}px 0 ${baselineSpacing}px 0;
```

src/components/ArticlePage/MarkdownHtml.js
```diff
@@ -89,8 +89,15 @@ const Html = styled.div`
...
     max-width: ${props => props.theme.contentMaxWidth}px;
+    border-top: 1px solid #CCCCCC;
+    padding-top: 29px;
```
kpuputti: This value is quite magical. You could write it in a more understandable way:

```suggestion
    padding-top: ${5 * baselineSmall - 1}px; // padding - border width
```

src/components/ArticlePage/MarkdownHtml.js
```diff
@@ -89,8 +89,15 @@ const Html = styled.div`
...
+
+    @media (min-width: ${baselineBreakpoint}px) {
+      margin: ${2 * baselineSpacing - 2}px 0 ${baselineSpacing}px 0;
```
kpuputti: Maybe use a multiple of `baselineLarge` here for clarity as well. But where does the `2px` come from? The border width is `1px`, or is the padding value affecting this?

kpuputti: You should also adjust the `first-child` top margin, now the baseline breaks at least in the Flex intro article. The height of the `h2` is not a multiple of the baseline.

Screen Shot 2019-04-08 at 16 14 23

src/components/ArticlePage/MarkdownHtml.js
```diff
@@ -89,8 +89,15 @@ const Html = styled.div`
...
+      padding-top: 41px;
```
kpuputti: Maybe replace this magical value with the calculation where it comes from for clarity?

src/components/ArticlePage/MarkdownHtml.js
jannekoivistoinen: hmm, this is a tricky one because some of the articles don't have a paragraph between ingress and H2, but most of them do.

PR #152, opened 2019-05-08
Time-based plan required

rap1ds:
The wording that says that Flex API defaults to day-based plan *which is always available* is very confusing. The actual effect that the users will see if they don't provide time-based plan for time-based listing is that there is no availability. Although the "day-based plan that is always available" is internally correct, it's not the behaviour the users will see.


src/docs/guides/24-how-to-use-time-based-bookings/index.md
```diff
@@ -30,10 +30,12 @@ ask that this adjustment is made to your process.
...
-type need to have a time-based availability plan. If an availability
-plan is not provided for a listing, the Flex API defaults to a day-based
-plan that is always available. Therefore, a time-based plan is required
-for time-based listings.
+type need to have a time-based availability plan. If a time-based
+availability plan is not provided for a listing, the Flex API defaults
+to a day-based plan, which will not allow time-based
+bookings. Therefore, a time-based plan is required for time-based
+listings.
+
```
lyyder:
```suggestion

PR #164, opened 2019-07-17
Update index.md

thomas-rocca:


src/docs/guides/09-how-to-set-up-currency-in-ftw/index.md
```diff
@@ -84,3 +84,7 @@ there is one place in FTW where prices are calculated:
 `BookingDatesForm`. See the file
 [EstimatedBreakdownMaybe.js](https://github.com/sharetribe/flex-template-web/blob/master/src/forms/BookingDatesForm/EstimatedBreakdownMaybe.js)
 to see how the Decimal.js library is used.
+
+## Your Saunatime demo listings 
+
+If you just started, you likely have a few demo Saunatime listings in your Console. These listings use EUR as their currency. Configuring a currency different than EUR will result in an error as the FTW cannot convert the currency of existing listings. You can ignore this error by closing these listings in your Console and creating new listings. Alternatively, you can use the [Marketplace API](https://www.sharetribe.com/api-reference/#update-listing) to change the currency of these listings.
```
OtterleyW: I just realized that actually updating the demo listings is not possible via Martketplace API because that requires authenticating as a the author of the listing and it's not possible. 

So this part "Alternatively, you can use the [Marketplace API](https://www.sharetribe.com/api-reference/#update-listing) to change the currency of these listings." should be removed. 

Maybe it would be also worth mentioning that the closed listing are not visible in the search page (so that's why the easiest solution is just to close the listings. Other solution could be to ban the demo users because then the listings will be deleted.

PR #168, opened 2019-08-05
Add background article about automatic payments in tx process

bladealslayer:
Illustrates usage of saved payment cards and off-session payments in tx
process.


src/docs/background/16-off-session-payments/index.md
```diff
@@ -0,0 +1,64 @@
...
+throughout the booking period.
+
+## Transaction process example
+
+Suppose your sauna rentals marketplace should allow customers to book saunas up
+to a year in advance, but the customer is charged only a certain time in advance
+before the booking. The figure below illustrates how a part of your transaction
+process might look like.
+
+![Example transaction process with delayed payment](tx-delayed-payments.png 'Example transaction process with delayed payment')
```
rap1ds: Comment: In the example process, the "pending-payment" is the state where the client app should prompt the customer to pay, right? But that's also the same state where we try to do the automatic payment transition. So, doesn't this lead to potential race condition? E.g. the user happens to open the app right at the same time when the transition "mark-pending-payment" completes and then he tries to make a payment in the UI, but that transition will fail because automatic "auto-payment" transition is running and locking the transition.

Would it make a bit more robust if the transitions were like this:

- From `accepted` via `auto-payment` to `paid`, 30 days and 10 minutes before booking starts
- From `accepted` via `mark-pending-payment` to `pending-payment`, 30 days before booking starts

In practice it would change the process so that `mark-pending-payment` state is only reached when the automatic payment fails.

bladealslayer: @rap1ds you are correct, there is a race condition. However, since `mark-pending-payment` is happening when the customer isn't necessarily on the website, I think the risk is small.

What you suggest is unfortunately not possible in Tegel at the moment: what you described is two delayed transitions leaving the same state. You can currently define those like that, but the execution in the engine is that both times are computed and *only the earlier* transition is scheduled. I.e. right now you can't use second delayed transition is a poor mans workaround for "if ... else ...". We discussed implementing this capability briefly with @ovan but we weren't sure allowing multiple delayed transitions is the right solution here in the first place.

PR #181, opened 2019-10-02
Custom components

rap1ds:
This PR adds support for custom components and implements Asciinema component


src/components/Asciinema.js
```diff
@@ -0,0 +1,38 @@
...
+    this.ref = React.createRef();
+  }
+
+  componentDidMount() {
+    const script = document.createElement('script');
+
+    script.setAttribute('id', 'asciicast-' + this.props.id);
+    script.setAttribute(
+      'src',
+      'https://asciinema.org/a/' + this.props.id + '.js'
```
kpuputti: Could the `id` prop have some other name like `session-id` etc.? Mostly because the `id` attribute has a special meaning in HTML, and it would yield invalid HTML if the same `id` would appear more than once in a page.

src/docs/tutorials/01-getting-started-with-ftw/index.md
```diff
@@ -45,6 +45,8 @@ basic development tooling:
 
 ## Setup FTW
 
+<asciinema id="267480"></asciinema>
+
```
kpuputti: This probably should be removed, it's the FTW getting started article.

PR #188, opened 2019-10-09
Update email template docs

lyyder:
- Update context description in email templates reference article
- Add an explanation on accessing email context properties to the Edit templates with CLI tutorial


src/docs/references/03-email-templates/index.md
```diff
@@ -331,22 +331,28 @@ Context for transaction emails:
...
+    "private-data": "extended-data",
+    "public-data": "extended-data",
+    "protected-data": "extended-data"
   },
   "marketplace": {
     "name": "string",
-    "url" "string"
+    "url": "string"
   },
   "recipient-role": {"one-of": ["provider", "customer"]},
```
kpuputti: Could you change these `one-of` things? I think the syntax is really weird and could be a lot simpler:

```suggestion
   "recipient-role": "provider", // either "provider" or "customer"
```

src/docs/references/03-email-templates/index.md
```diff
@@ -331,22 +331,28 @@ Context for transaction emails:
...
-    "display-name": "string"
+    "display-name": "string",
+    "private-data": "extended-data",
+    "public-data": "extended-data",
+    "protected-data": "extended-data"
   },
   "transaction": {
     "id": "uuid",
-    "line-items": [
+    "tx-line-items": [
```
kpuputti: Has this changed or was the reference wrong?

lyyder: @kpuputti It has changed.

src/docs/references/03-email-templates/index.md
lyyder: @kpuputti or what about
```
"recipient-role": "string", // either "provider" or "customer"
```
?

lyyder: Changing to the above to get this forward.

src/docs/references/03-email-templates/index.md
kpuputti: Ok, good to know, just made sure. I assume all the deprecation, possible braking changes and other stuff is taken into account.

lyyder: Yep, so `line-items` is still available in the context but not documented anywhere.

PR #194, opened 2019-10-20
Update default process alias

lyyder:
Change default process alias references to the one used with the default
process for new marketplaces.


lyyder:
@bladealslayer yes, decided that a more thorough thinking process is required before deciding on the example process names.

PR #226, opened 2020-02-14
Login as user how-to

lyyder:


src/docs/guides/28-how-to-enable-login-as-user/index.md
```diff
@@ -0,0 +1,289 @@
...
+published: true
+---
+
+The _Login as user_ feature allows marketplace operators to log into
+their marketplace as a specific user of the marketplace. This helps
+operators to experience their marketplace as their users do and to find
+out what is wrong when their users are reporting problems. The feature
+also comes in handy when a marketplace user asks for help with managing
+their data and listings.
+
```
bladealslayer: I'd state already here briefly that taking this into use requires action, unless you're using latest FTW.

I would highlight clearly here also the limitations of what you can not do when using this (modify Stripe account details, send messages, initiate or transition transactions). Now that's mentioned at the end of a very technical paragraph, which non-technical operators might just skip.

src/docs/guides/28-how-to-enable-login-as-user/index.md
```diff
@@ -0,0 +1,289 @@
...
+As context, here's a quick description of the technical implementation
+of how the Login as user works to make it easer to understand the
+changes it requires. The authentication flow uses the _authorization
+code_ grant type defined in the OAuth2. Console works as an
+_authorization server_ that issues an authorization code for FTW. FTW
+then uses this code to obtain an access token from Auth API. The access
+token is valid for 30 minutes and it does not come with a refresh token.
+The token can be used as a normal token obtained with a password login
+excluding updating payment information, sending messages, or updating
```
bladealslayer: s/updating/initiating or transitioning/ to be more clear?

src/docs/guides/28-how-to-enable-login-as-user/index.md
```diff
@@ -0,0 +1,289 @@
...
+      code_verifier: codeVerifier,
+    })
+    .then(() => res.redirect('/'))
+    .catch(() =>
+      res.status(401).send('Unable to authenticate as a user')
+    );
+});
+
+module.exports = router;
+```
```
bladealslayer: How about if we link to the actual files in the FTW repos here and say that up-to-date source code can be found there? In case we tweak those and forget to update the code on this page.

src/docs/guides/28-how-to-enable-login-as-user/index.md
```diff
@@ -0,0 +1,289 @@
...
+
+```javascript
+app.use('/api', apiRouter);
+```
+
+Now you can log into the marketplace from Console's user view.
+
+### Show a banner when logged in as a user
+
+The endpoints of the previous chapter enable the feature. However, this
```
bladealslayer:
```suggestion
The endpoints of the previous section enable the "login as" feature. However, this
```

src/docs/guides/28-how-to-enable-login-as-user/index.md
lyyder: Good point. Up till now we have pasted the contents to the docs articles. I don't know which is approach is more clear to majority of users. I'll keep this like this for now and we can discuss this further if we want to update the guide.

PR #235, opened 2020-02-27
Remove guidance about requesting access

lyyder:
With automated onboarding there's no need to request access but to
create a marketplace.


src/docs/background/09-how-to-launch/index.md
```diff
@@ -29,12 +29,13 @@ to do that now.
...
-in to [Flex Console](https://flex-console.sharetribe.com) (the "admin
-section" of your Flex marketplace). If you don't have access yet, you
-can request access by
-[filling out this form](https://sharetribe.typeform.com/to/BI9M4O). If
-you have gotten access to Flex but have forgotten your credentials,
+access a Flex marketplace. This means that you have an account in
+[Flex Console](https://flex-console.sharetribe.com) (the "admin section"
+of your Flex marketplace). If you don't have an account yet, you can
+create one in
+[the Sharetribe Flex website](https://www.sharetribe.com/products/flex/).
```
kpuputti: We can also point directly to the Flex creation wizard in the modal with a hash URL. Not sure if it will change still as things are not fully finished, but at the moment the URL would be https://www.sharetribe.com/#start-building-with-flex

src/docs/tutorials/01-getting-started-with-ftw/index.md
```diff
@@ -42,10 +42,11 @@ the
...
 article.
 
-## Request API access
+## Create a marketplace
 
-To use the Marketplace API, you will need a client ID. You can request
-access at https://sharetribe.typeform.com/to/BI9M4O
+To use the Marketplace API, you will need a client ID. You can obtain
+one by creating a new Flex marketplace at
+[the Sharetribe Flex website](https://www.sharetribe.com/products/flex/).
```
kpuputti: Same here, you could link directly to the form.

PR #271, opened 2020-03-31
Add desktop search to Flex docs

lyyder:
Adds desktop search to the docs. The search used Algolia Docsearch.

The Docsearch is integrated using the [gatsby-plugin-algolia-docsearch plugin](https://www.gatsbyjs.org/packages/gatsby-plugin-algolia-docsearch/). The plugin uses Gatsby's `onRenderBody` API to attach required Docsearch code snippets and `MutationObserver` browser API to recognize when Docsearch input element is available and can be initiated.


src/components/Topbar/Search.js
```diff
@@ -1,17 +1,35 @@
...
   height: 60px;
 
   @media (min-width: ${baselineBreakpoint}px) {
-    // Same width as the logo in the left for symmetry
-    width: 120px;
-    height: 72px;
+    width: 220px;
+    height: 40px;
+    border: 1px solid ${props => props.theme.searchBorderColor};
+    border-radius: 20px;
```
Gnito: Remove the border from here and and add it directly to input.

Input could contain something like:
```css
  width: 220px;
  height: 40px;
  border: 1px solid #d1d1d1;
  border-radius: 20px;
  padding: 0 10px 0 20px;
  font-size: 16px;
```

Gnito: To the **src/components/GlobalStyles.js**, you could add something like this:
```css
  .algolia-autocomplete {
    ${fonts['CircularStd-Book'].styles}

    & .algolia-docsearch-suggestion--title {
      ${fonts['CircularStd-Bold'].styles}
    }
  }
```

After button style-rules.

lyyder: > => i.e remove the border from here. You can add it directly to Input.

How about the magnifying glass, should it be outside the input, so that there would be a container element that holds both?

lyyder: Probably the icon can also go into the input, having the border in the input instead of a wrapping `div` does feel more natural indeed.

PR #299, opened 2020-05-11
Remove article reading time estimation feature

kpuputti:
This PR removes the reading time estimation as it hasn't been used at all.


Gnito:
Removing the reading time is OK for me 👍

PR #302, opened 2020-06-01
Update login as docs with changed API router location in the repo

kpuputti:
This PR updates the docs to match the API router setup refactored here: https://github.com/sharetribe/ftw-daily/pull/1294


src/docs/cookbook-manage/enable-login-as-user/index.md
```diff
@@ -78,179 +78,10 @@ const authenticated = authInfo =>
...
-    .catch(() =>
-      res.status(401).send('Unable to authenticate as a user')
-    );
-});
-
-module.exports = router;
-```
+add them to the Node server, copy
+[server/apiRouter.js](https://github.com/sharetribe/ftw-daily/blob/master/server/apiRouter.js),
+`server/api/initiate-login-as.js`, and `server/api/login-as.js` files
```
OtterleyW, line 83: I think it would be nice if these two files were also links to FTW-repo

kpuputti: I wanted to avoid direct links to files since those are bound to change in the long term. But it doesn't really make sense to have one file linked and two others not. So you have a point. Would it work if we only linked to the `server/` directory and then mentioned the file names that should be changed?

OtterleyW: @kpuputti if we want to avoid having direct links to files then linking to the directory would be a good solution. I don't have strong opinion on which one we should pick but thought it feels a little weird that one file is linked and other ones are not...

PR #309, opened 2020-06-23
Add privileged-set-line-items action doc and deprecate others

sktoiva:
Direct link to the action reference preview: https://deploy-preview-309--sharetribe-flex-docs-site.netlify.app/docs/references/transaction-process-actions/

### DIFF
--- a/src/docs/references/transaction-process-actions/index.md
+++ b/src/docs/references/transaction-process-actions/index.md
@@ -45,7 +45,67 @@ Initialize a new transaction from a listing.
 
 **Configuration options**: -
 
+### :action/privileged-set-line-items
+
+Defines transaction price and breakdown. Sets given line items and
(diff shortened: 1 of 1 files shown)


src/docs/references/transaction-process-actions/index.md
```diff
@@ -45,8 +45,70 @@ Initialize a new transaction from a listing.
 
 **Configuration options**: -
 
+### :action/privileged-set-line-items
+
+Enables custom pricing. Sets given line items and calculates totals for
+each line item and for the transaction.
```
lyyder, line 51: Suggestion: Instead of _Enables custom pricing._ let's write _Defines transaction price and breakdown._ or something of the sort. This action is now the default way to set the pricing for a marketplace so in that sense it does not enable any special feature but is used for the standard pricing in Flex.

PR #312, opened 2020-06-26
Update extended data note

kpuputti:


src/docs/references/extended-data/index.md
```diff
@@ -16,9 +16,9 @@ extended data for some value types.
 
 ## Types of extended data
 
-**Note:** Extended data support in API resources is work in progress.
-See the [API reference](/references/api/) for each resource for
-information on supported extended data.
+**Note:** Some API resources don't have all extended data fields
+implemented. See the [API reference](/references/api/) for each resource
+for information on supported extended data.
```
lyyder, line 21: This is good but I get the feeling that we're planning to implement all fields on every resource, but have not done it yet, which still communicates that this is WIP. We could even say that _Available extended data fields vary between API resources. See the..._ or something to communicate that not every field is available in every resource but that this if a fully functional Flex feature. However, this is not a super strong opinion so feel free to keep the current message if think it works better.

Gnito:
@kpuputti nice small change. Should this be merged?

kpuputti:
@Gnito Yeah, I left it open after the review comment. I've been pushing this to be a cooldown task, but never got to it. Maybe someone that has a good overall picture of what things we are planning to implement and how that should be communicated could define what to put here.

PR #315, opened 2020-07-08
Transaction metadata and operator transitions via Integration API

bladealslayer:


bladealslayer:
@sktoiva good question. There is no inherent restriction, as long as the client would be able to supply the required transition parameters. So, technically it's possible to design a process where all but the initial transitions are operator ones and all calls happen via customer's backend (perhaps in response to UI interaction of the provider/customer and UI calling API in their backend). Not sure whether I'd be recommending building such process, but it is possible.

Any suggestion where to make the clarification that all actions are allowed?

bladealslayer:
PR got somehow bugged and doesn't show all commits. Will open new one.

PR #316, opened 2020-07-10
Add doc about removing Stripe and payments

OtterleyW:


src/docs/cookbook-payments/removing-stripe/index.md
```diff
@@ -0,0 +1,332 @@
...
+  customizations.
+published: true
+---
+
+# Removing Stripe
+
+In this guide, we will go through the minimum changes needed to make the booking
+flow work without Stripe. Most likely you need to customize some parts (e.g.
+transaction process, email templates, checkout page) more depending on your
+marketplace idea. You might also want to remove e.g. unused Stripe components
```
Gnito, line 18:
```suggestion
marketplace idea. For example, you might also want to remove unused Stripe components
```

Gnito: Or, "E.g. [...]"

src/docs/cookbook-payments/removing-stripe/index.md
```diff
@@ -0,0 +1,332 @@
...
+- :action/stripe-confirm-payment-intent
+- :action/stripe-create-payout
+- :action/stripe-refund-charge
+- :action/stripe-refund-payment
+
+If you are using the FTW default process with strong customer authentication
+(e.g. flex-default-process) you need to remove confirming the payment.
+Otherwise, the transaction will get stuck. The simplest way to do this is to
+remove `pending-payment` and `expire-payment` states and point `request-payment`
+and `equest-payment-after-enquiry` directly to `preauthorized` state. However,
```
Gnito, line 36:
```suggestion
and `request-payment-after-enquiry` directly to `preauthorized` state. However,
```

src/docs/cookbook-payments/removing-stripe/index.md
```diff
@@ -0,0 +1,332 @@
...
+when you are building the transaction process to your marketplace it might make
+sense to rename some of the states and transitions (e.g. preauthorized) so that
+they make more sense in your use-case.
```
Gnito, line 39:
```suggestion
sense to rename some of the states and transitions so that
they make more sense in your use-case. E.g. *preauthorized* state refers to the state of payment - it is preauthorized to be captured.
```

src/docs/cookbook-payments/removing-stripe/index.md
```diff
@@ -0,0 +1,332 @@
...
+In
+[`EditListingWizard.js`](https://github.com/sharetribe/ftw-daily/blob/master/src/components/EditListingWizard/EditListingWizard.js)
+we are checking if the provider has a Stripe account with all the required
+information before we allow them to publish listings. This check is done in the
+`handlePublishListing` function and needs to be removed.
+
+```js
+  handlePublishListing(id) {
+    const { onPublishListingDraft} = this.props;
+    onPublishListingDraft(id);
```
Gnito, line 92:
```suggestion
    this.props.onPublishListingDraft(id);
```

And actually, this whole function could be removed, but that would require another line to be edited (the one where this.handlePublishListing(id) is called. For the sake of this guide, it's probably better to keep the change inside this one method.

src/docs/cookbook-payments/removing-stripe/index.md
```diff
@@ -0,0 +1,332 @@
...
+  }
+```
+
+After changing the `handlePublishListing` you can remove unused code. You also
+might want to clean up `EditListingPage` and remove the Stripe related props we
```
Gnito, line 97: Is there any link to the correct lines (where to begin this cleaning) or is there too many of them?

src/docs/cookbook-payments/removing-stripe/index.md
```diff
@@ -0,0 +1,332 @@
...
+payment method have added a lot of logic to the current page. We can use the
+[edited version of CheckoutPage from FTW 2.17.1](/tutorial-assets/SimpleCheckoutPage.js)
+as a simpler starting point.
+
+Some of the functions on later versions of CheckoutPage.duck.js has changed so
+here are the main changes done in the edited code:
+
+- In `handleSubmit` function we only call onInitiateOrder function which
+  internally decides if the transaction should be initiated or transitioned. We
+  also need to handle sending the possible inital message after that.
```
Gnito, line 121:
```suggestion
  also need to handle sending the possible initial message after that.
```

src/docs/cookbook-payments/removing-stripe/index.md
```diff
@@ -0,0 +1,332 @@
...
+will remind users to create a Stripe account. The same modal is used for
+reminding about email verification so we should just remove the Stripe related
+code from the component.
+
+###4. Edit CheckoutPage
+
+Removing Stripe from `CheckoutPage` is probably the most complicated task in
+this list because the Strong Customer Authentication (SCA) and ability to save
+payment method have added a lot of logic to the current page. We can use the
+[edited version of CheckoutPage from FTW 2.17.1](/tutorial-assets/SimpleCheckoutPage.js)
```
Gnito, line 113: Don't hide a new resource as a link inside paragraph. It's easier to spot these if they are more prominent.

----

We can use the edited version of CheckoutPage from FTW 2.17.1 as a simpler starting point:

- [SimpleCheckoutPage.js](/tutorial-assets/SimpleCheckoutPage.js)

src/docs/cookbook-payments/removing-stripe/index.md
```diff
@@ -0,0 +1,332 @@
...
+Remove the actions related to payments e.g. in flex-default-process
+:action/privileged-set-line-items and :action/calculate-full-refund. In older
+processes this means actions related to calculating total price, commission etc.
+Read more about actions from
+[transaction process actions refrence article.](/references/transaction-process-actions/)
+Remember also to update email templates which contain pricing information.
+
+### Edit API calls in FTW backend
+
+If you are using FTW-daily v6.0.0 or higher or FTW-hourly v8.0.0 or higher you
```
Gnito, line 314: Lot's of "or" conjunctions that are connecting items a bit strangely. 

One alternative suggestion:

```suggestion
If you are using FTW-daily v6.0.0 or higher (or FTW-hourly v8.0.0 or higher) you
```

src/docs/cookbook-payments/removing-stripe/index.md
```diff
@@ -0,0 +1,332 @@
...
+  In some cases, you might want to remove Stripe integrations or even all the
+  payments from FTW. This article gives you a starting point for these
+  customizations.
+published: true
+---
+
+# Removing Stripe
+
+In this guide, we will go through the minimum changes needed to make the booking
+flow work without Stripe. Most likely you need to customize some parts (e.g.
```
Gnito, line 16: I think this should include a bit more about what to expect. This guide is very rough by nature - clients clearly need to go through lots of (unguided) changes on several files. So, this paragraph could prepare devs & admins that going through these changes is going to be a relatively big customization project.


 FTW templates are made for a specific transaction process.
- Payments are a core feature that we have tried to cover with these provided templates.

```suggestion
flow work without Stripe. This is not an easy customization project. FTW templates are made for a specific transaction process and payments are a core feature: its ubiquitous nature means that you need to touch quite many components and files. Most likely you need to customize some parts (e.g.
```

src/docs/cookbook-payments/removing-stripe/index.md
OtterleyW: On `EditLisitngPage` there are mostly props that are passed forward and some variables related to Stripe so not very easy to point to just one line.

PR #319, opened 2020-07-14
Group actions in reference doc

Gnito:
Group actions under subheadings:

Screenshot 2020-07-14 at 16 02 27

This is an initial improvement for the transaction process actions reference doc. ToC shows 2 levels of headings and actions list is too long for it.

We might want to add a full list of actions to the beginning of the article later. I.e. we could explicitly write that list or create component that automatically injects those links.

### DIFF
--- a/src/docs/references/transaction-process-actions/index.md
+++ b/src/docs/references/transaction-process-actions/index.md
@@ -25,7 +25,9 @@ reference article.
 
 ## Actions
 
-### :action.initializer/init-listing-tx
+### Transaction initialization
+
+#### :action.initializer/init-listing-tx
 
 **NOTE**: this action is implicit and must not be in the process
 description.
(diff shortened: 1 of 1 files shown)


src/docs/references/transaction-process-actions/index.md
```diff
@@ -25,7 +25,9 @@ reference article.
 
 ## Actions
 
-### :action.initializer/init-listing-tx
+### Built-in
```
bladealslayer, line 28: How about "Transaction initialization"? That communicates better the purpose of the action(s) here.

src/docs/references/transaction-process-actions/index.md
```diff
@@ -581,7 +589,9 @@ Action to publish any reviews in the transaction.
 
 **Configuration options**: -
 
-### :action/reveal-customer-protected-data
+### Protected data
```
bladealslayer, line 592: I'd call this "Extended data". In my open PR there's already an action that deals with metadata. Maybe no need for separate heading for each type of extended data?

bladealslayer:
Closing in favor of #320 @Gnito

PR #322, opened 2020-07-22
Create a new transaction process

Gnito:
Process alias: *cottagedays-daily-booking/release-1*


src/docs/tutorial-transaction-process/create-transaction-process/index.md
```diff
@@ -0,0 +1,411 @@
...
+        [TRANSITION_EXPIRE]: STATE_DECLINED,
+        [TRANSITION_ACCEPT]: STATE_ACCEPTED,
+      },
+    },
+
+    [STATE_DECLINED]: {},
+    [STATE_DECLINED_BY_OPERATOR]: {},
+```
+
+**Note**: You could also update the id of the graph.
```
OtterleyW, line 317: I think this could be a separate step before the current step 3 or then inside step 3 but mentioned before updating the actual graph (because the id comes in the code first).

src/docs/tutorial-transaction-process/create-transaction-process/index.md
```diff
@@ -0,0 +1,411 @@
...
+marketplace.
+
+In addition, we also add a new decline transition for the operator and
+update the client app so that it uses the new process.
+
+### Clone Flex example processes repository
+
+Writing _process.edn_ file and email templates from scratch is a bit
+tedious task. We'll make our life a bit easier by cloning Flex example
+processes repository:
```
OtterleyW, line 27: I think we could mention more clearly already here which process we are going to use as a starting point. Then we could move the extra info about preauth-daily-booking process already to this section.

Maybe we could even add a link to the example process repo and explain briefly what kind of processes there are. Though most of them are using deprecated pricing actions so maybe that's not very useful after all.

src/docs/tutorial-transaction-process/create-transaction-process/index.md
Gnito: It's not strictly speaking necessary to update it, but I'll mention this at the beginning of step 3.

PR #323, opened 2020-07-23
Customize pricing tutorial

OtterleyW:


src/docs/tutorial-transaction-process/customize-pricing/index.md
```diff
@@ -0,0 +1,469 @@
...
+In _EditListingPricingPanel_ we need to edit the `onSubmit` function to save the
+new public data field `cleaningFee`. Because we are using `FieldCurrencyInput`
+in this example as the input, the `cleaningFee` will be Money object when we get
+that from `values`. Money object can't be used directly as public data so we
+need to create a JSON object with keys `amount` and `currency` and use the JSON
+object in the API call.
+
+_Note:_ `price` attribute is one of the listing's default attributes so it's
+passed to Marketplace API directly unlike the new public data `cleaningFee`
+which needs to be under the `publicData` key.
```
Gnito, line 42: Näissä noteissa mulla on ollut yhtenäisyyden nimissä tämmöinen notaatio:

```suggestion
> **Note**: `price` attribute is one of the listing's default attributes so it's
> passed to Marketplace API directly unlike the new public data `cleaningFee`
> which needs to be under the `publicData` key.
```

src/docs/tutorial-transaction-process/customize-pricing/index.md
```diff
@@ -0,0 +1,469 @@
...
+/>
+...
+```
+
+After adding the new translations keys and some padding to price input in CSS
+file, the EditListingPricingPanel should look something like this:
+
+![EditListingPricePanel](./editlistingpricepanel.png)
+
+# Update BookingDatesForm
```
Gnito, line 119: There can't be another h1 in the article. `##`

src/docs/tutorial-transaction-process/customize-pricing/index.md
```diff
@@ -0,0 +1,469 @@
...
+and you can move forward to the next section: Add a transaction line item for
+the cleaning fee.
+
+### Prepare props
+
+The cleaning fee is now saved to listing's public data so we can find it under
+the `publicData` key from listing's attributes. Because we decided that adding a
+cleaning fee to listing's information is optional, we need to check if the
+cleaningFee exists in public data or not.
+
```
Gnito, line 137: I think the file location would make sense here. I was not sure where this was supposed to be.

src/docs/tutorial-transaction-process/customize-pricing/index.md
```diff
@@ -0,0 +1,469 @@
...
+      this.props.onFetchTransactionLineItems({
+        bookingData: { startDate, endDate, hasCleaningFee },
+        listingId,
+        isOwnListing,
+      });
+    }
+  }
+```
+
+# Add a new line-item for the cleaning fee
```
Gnito, line 348: There can't be more than one h1 in the article. `##`

src/docs/tutorial-transaction-process/customize-pricing/index.md
OtterleyW: Good catch! It seems to be missing one paragraph about BookingPanel but I added that now.

src/docs/tutorial-transaction-process/customize-pricing/index.md
```diff
@@ -0,0 +1,482 @@
...
+for step-by-step instructions on how to add public data field for listing.
+
+### Save to public data
+
+In _EditListingPricingPanel_ we need to edit the `onSubmit` function to save the
+new public data field `cleaningFee`. Because we are using `FieldCurrencyInput`
+in this example as the input, the `cleaningFee` will be Money object when we get
+that from `values`. Money object can't be used directly as public data so we
+need to create a JSON object with keys `amount` and `currency` and use the JSON
+object in the API call.
```
Gnito, line 38: This is a bit personal reference, but these yellow code-block highlights look pretty ugly. I've started to reduce the amount of code-block highlights in favour of italics and bolded words. Could you check if this would look better:


```suggestion
In _EditListingPricingPanel_, we need to edit the _onSubmit_ function to save the
new public data field called **cleaningFee**. Because we are using _FieldCurrencyInput_ component
in this example as the input of choice, the _cleaningFee_ variable will be Money object when we get
that from the submitted _values_. Money object can't be used directly as public data so we
need to create a JSON object with keys **amount** and **currency** and use it in the underlying API call.
```

src/docs/tutorial-transaction-process/customize-pricing/index.md
```diff
@@ -0,0 +1,482 @@
...
+```diff
+    ...
+    lineItems,
+    fetchLineItemsInProgress,
+    fetchLineItemsError,
++   cleaningFee,
+  } = fieldRenderProps;
+```
+
+We want show the amount of cleaning fee to the user in the checkbox label so we
```
Gnito, line 233:
```suggestion
We want to show the amount of cleaning fee to the user in the checkbox label so we
```

src/docs/tutorial-transaction-process/customize-pricing/index.md
```diff
@@ -0,0 +1,482 @@
...
+![Cleaning fee checkbox](./cleaningFeeCheckbox.png)
+
+### Update the bookingData
+
+Next, we want to pass the value of the cleaning fee checkbox forward as part of
+the `bookingData`. This is needed so that we can show the selected cleaning fee
+as a new row on the booking breakdown. To achieve this, we need to edit the
+`handleOnChange` function which takes the values from the form and calls the
+`onFetchTransactionLineItems` function for constructing the transaction line
+items. These line itemas are shown inside the `bookingInfoMaybe` component under
```
Gnito, line 306:
```suggestion
items. These line items are shown inside the `bookingInfoMaybe` component under
```

src/docs/tutorial-transaction-process/customize-pricing/index.md
```diff
@@ -0,0 +1,482 @@
...
+file, the EditListingPricingPanel should look something like this:
+
+![EditListingPricePanel](./editlistingpricepanel.png)
+
+## Update BookingDatesForm
+
+In our example the cleaning fee is optional and users can select that as an
+add-on to their booking. In this section, we will add the UI component for
+selecting the cleaning fee and pass the information about the user's choice to
+the FTW-backend.
```
Gnito, line 124: "FTW-backend" sounds confusing to me. We have FTW-daily and FTW-hourly - is the FTW-backend a new repository?

- the backend of our client app.
- the backend of FTW template.
- the backend of FTW-daily.

Gnito: Change the other 6 instances too.

src/docs/tutorial-transaction-process/customize-pricing/index.md
```diff
@@ -0,0 +1,482 @@
...
+fetch the pricing information like the cleaning fee amount directly from
+Marketplace API in FTW-backend, we can avoid this security risk.
+
+You can read more about privileged transitions from
+[privileged transitions background article](/background/privileged-transitions/).
+
+</extraInfo>
+
+When we want to add a new line item for cleaning fee, we'll need to update the
+pricing logic in the `/server/api-util/lineItems.js` file:
```
Gnito, line 393: No need to repeat the path.

```suggestion
pricing logic in the _lineItems.js_ file:
```

src/docs/tutorial-transaction-process/customize-pricing/index.md
```diff
@@ -0,0 +1,482 @@
...
+
+```shell
+└── server
+    └── api-util
+        └── lineItems.js
```
Gnito, line 398: Since also _lineItemHelpers.js_ is modified, you could add that too.
```suggestion
        ├── lineItems.js
        └── lineItemHelpers.js
```

src/docs/tutorial-transaction-process/customize-pricing/index.md
```diff
@@ -0,0 +1,482 @@
...
+  : null;
+
+const cleaningFeeLabel = intl.formatMessage(
+  { id: 'BookingDatesForm.cleaningFeeLabel' },
+  { fee: formattedCleaningFee }
+);
+```
+
+We will also add new translation key `BookingDatesForm.cleaningFeeLabel` to
+en.json file where we can use `fee` variable to show the price.
```
Gnito, line 250: Voisi vähän korostaa tuota en.jsonia, vaikka tähän ei alkaisikaan lisäämään polkua siihen.
```suggestion
We will also add new translation key _BookingDatesForm.cleaningFeeLabel_ to
***en.json*** file where we can use **fee** variable to show the price.
```

src/docs/tutorial-transaction-process/customize-pricing/index.md
```diff
@@ -0,0 +1,482 @@
...
+  <FieldCheckbox
+    id="cleaningFee"
+    name="cleaningFee"
+    label={cleaningFeeLabel}
+    value="cleaningFee"
+  />
+) : null;
+```
+
+Then we can add the `cleaningFeeMaybe` to the `Form` component
```
Gnito, line 272:
```suggestion
Then we can add the **cleaningFeeMaybe** to the returned `<Form>` component.
```

src/docs/tutorial-transaction-process/customize-pricing/index.md
OtterleyW: I changed the higlights to mostly italics now. I'm not sure what would be the best logic to choose between bolding or italics. Do you have any "rules" for this?

Maybe a better solution would be to discuss with Janne about changing the highlight formatting so that it would look nicer. I think it has been a common convention to use the highlight for variable, component, and file names quite some time already (or at least I have been doing that) so it might be confusing that we suddenly change this.

src/docs/tutorial-transaction-process/customize-pricing/index.md
OtterleyW: 👍  I changed that to "the backend of our client app" instead

src/docs/tutorial-transaction-process/customize-pricing/index.md
Gnito: @OtterleyW The highlight colour could be changed, but still I think the way we (and especially me) have used it is bad. The plethora of code-blocks makes it difficult to read the text. I still use it, but I've tried to minimize its usage in new articles to code expressions (not single words) and only max ~one per paragraph.

The problem is that code-block is not made to highlighting text, but to separate code statements (with ifs, elses and other English words) from a text that is part of a readable sentence. For example, assistive text-to-speech services probably can't make any sense of paragraphs where we have used code-blocks that much.

PR #324, opened 2020-07-24
Add new email notification article

OtterleyW:


src/docs/tutorial-transaction-process/add-email-notification/index.md
```diff
@@ -0,0 +1,250 @@
...
+- _new-booking-request-html.html_ - contains the template for the HTML
+  version of the mail
+
+We need to rename these files to:
+
+- _new-booking-request-for-customer-html.html_
+- _new-booking-request-for-customer-subject.txt_.
+
+Remember to follow the naming convention where the file ends with
+\*-html.html or \*-subject.txt. Otherwise, the notification files will
```
Gnito, line 68: Does this work with italics?

```suggestion
_\*-html.html_ or _\*-subject.txt_. Otherwise, the notification files will
```

OtterleyW:
@Gnito when looking at the artice again I agree that not using extraInfo would be better because it hides quite important part. I also noticed there was some unused css so I removed that and added a simpler css example there.

PR #336, opened 2020-09-07
install circular and fix the readability issue on the sidebar

jannekoivistoinen:
This PR;
- Adds Circular font back to the Flex Docs 
- Improves sidebar's readability with darkened and bolded text

Before; 
image

After;
image

### DIFF
--- a/src/font-faces.css
+++ b/src/font-faces.css
@@ -69,6 +69,7 @@ EULA          https://lineto.com/assets/pdf/EULA.pdf
   src: url("https://assets-sharetribecom.sharetribe.com/webfonts/circular/CircularXXSub-ThinSubset.woff2") format("woff2");
   font-weight: 100;
   font-style: normal;
+  font-display: fallback;
 }
 
 @font-face {
@@ -76,6 +77,7 @@ EULA          https://lineto.com/assets/pdf/EULA.pdf
   src: url("https://assets-sharetribecom.sharetribe.com/webfonts/circular/CircularXXSub-ThinItalicSubset.woff") format("woff");
   font-weight: 100;
   font-style: italic;
+  font-display: fallback;
 }
 
 @font-face {
@@ -83,6 +85,7 @@ EULA          https://lineto.com/assets/pdf/EULA.pdf
   src: url("https://assets-sharetribecom.sharetribe.com/webfonts/circular/CircularXXSub-ThinItalicSubset.woff2") format("woff2");
   font-weight: 100;
   font-style: italic;
+  font-display: fallback;
 }
 
 @font-face {
@@ -90,6 +93,7 @@ EULA          https://lineto.com/assets/pdf/EULA.pdf
   src: url("https://assets-sharetribecom.sharetribe.com/webfonts/circular/CircularXXSub-LightSubset.woff") format("woff");
   font-weight: 200;
   font-style: normal;
+  font-display: fallback;
 }
 
 @font-face {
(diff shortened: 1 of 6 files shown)


src/font-faces.css
```diff
@@ -9,68 +9,274 @@
...
+
 @font-face {
-  font-family: CircularStd;
-  src: url("./font-files/circular-std/CircularStd-Black.woff2") format("woff2");
-  font-weight: 800;
+  font-family: "Circular";
+  src: url("https://assets-sharetribecom.sharetribe.com/webfonts/circular/CircularXXSub-ThinSubset.woff2") format("woff2");
+  font-weight: 100;
   font-style: normal;
-  font-display: fallback;
```
kpuputti, line 18: Is there a specific reason to remove the font display behavior? I remember spending time fine-tuning this based on some best practices. I understand that first-time flashing of fallbacks might be annoying, but there are accessibility reasons at play here.

src/font-faces.css
```diff
@@ -9,68 +9,274 @@
...
 }
 
 @font-face {
-  font-family: CircularStd;
-  src: url("./font-files/circular-std/CircularStd-BlackItalic.woff2") format("woff2");
-  font-weight: 800;
+  font-family: "Circular";
+  src: url("https://assets-sharetribecom.sharetribe.com/webfonts/circular/CircularXXSub-ThinItalicSubset.woff") format("woff");
```
kpuputti, line 76: Why do you add both the `woff2` and `woff` formats if only the first was enough previously? Is there some browser support that we were missing before? How does that affect what files browsers download?

kpuputti (changes requested):

src/font-faces.css
jannekoivistoinen: "Web Open Font Format 2 (WOFF2): WOFF2 is an update to the original WOFF format. Developed by Google, this is considered the best format of the bunch because it offers smaller file sizes and better performance for modern browsers that support it." -  https://css-tricks.com/understanding-web-fonts-getting/

So WOFF2 is an update to WOFF, but having WOFF as a backup to older browsers is still recommended.

src/font-faces.css
jannekoivistoinen: Thanks for this @kpuputti - I totally missed this by accident as I only copied the CSS from the font-provider and it didn't have the font-display: fallback in it. Totally new thing for me, I'll add it back. Good catch 🍬

src/font-faces.css
kpuputti: We can of course add support for older browsers, but WOFF2 support looks pretty good already: https://caniuse.com/?search=woff2

I'd personally go just with WOFF2 here as the support is good and the fallback behavior doesn't break anything. It simplifies the code, but feel free to decide otherwise if you think the browser support isn't enough.

jannekoivistoinen: I would like to keep the WOFF still in. All major font providers still include WOFF and WOFF2 files in their code and deliverables and WOFF2's global rate is ~95% and WOFF's is ~97.5%. Supporting both doesn't increase page load, or render times and the font-faces.css file is not something that we (hopefully) have to touch regularly.

To me, the pro (wider support) outweights the con (more code in font-faces.css files) and I would like to keep them both in. 

But if you'd like to simplify the code without sacrificing the support, we could remove the unused styles and weights from the code :)

kpuputti (approved):

PR #338, opened 2020-09-24
Propose & update bookings

bladealslayer:
A couple of articles still refer to the deprecated `create-booking`:

* https://www.sharetribe.com/docs/references/transaction-process-format/ this one is about the process format, so perhaps not so important that it uses several deprecated actions as examples.
* https://www.sharetribe.com/docs/flex-cli/edit-transaction-process-with-flex-cli/ is a how-to for Flex CLI. The example edits are outdated (how to change commission), so I didn't see value in updating only the mention of `create-booking` action.

Related API reference update: https://github.com/sharetribe/sharetribe-custom/pull/1873

TODO:
- [x] update action reference after changes in https://github.com/sharetribe/sharetribe-custom/pull/1858


src/docs/references/availability/index.md
```diff
@@ -128,18 +132,47 @@ For time-based plans, both availability exceptions and bookings are
...
+All bookings in `pending` or `accepted` states count as reservation
+against the listing's availability. On the other hand, bookings in the
+`proposed`, `canceled` or `declined` states do not affect the
+availability of the listing.
+
+The figure below illustrates the possibel booking states and transitions
+and the corresponding actions that you can use in your transaction
+process.
+
+![Booking states](./booking-states.png)
```
ovan, line 151: Comment: My small improvement idea here is to show the diagram without `:action/update-booking` transitions. They add a lot of noise but are not critical for understanding how the states work and how you transition between them. Simply mentioning that there's also ´:action/update-booking` which can be used to update times and display times but that does not affect state, would suffice in this context.

ovan (commented):
About the two things you mentioned leaving out.
1. https://www.sharetribe.com/docs/references/transaction-process-format/ - I would include updating this one to not use the deprecated action. On a quick look it didn't seem that details of create-booking are used anywhere so it should be as simple as updating the example edn to use `create-pending-booking`, or did I miss something? 
2. https://www.sharetribe.com/docs/references/transaction-process-format/ - I agree that updating the example case is not in this scope. This article has been in need of update already since privileged transitions and I hadn't even realized that. Still, I would just find and replace create-booking with create-pending-booking here to not have references to deprecated action that we no longer wish people to use. I don't see any harm doing that unless again I'm missing some subtlety?

src/docs/references/availability/index.md
bladealslayer: Good idea!

bladealslayer:
@ovan 👍 

1. There is a bit more, but not much. The action configuration comes up in one of the sections, so a bit of wording update is needed.

2. Simple replace would be enough here. It's just not enough for making the article make sense as a whole. Effectively it will contain a process example with combination of actions that are neither exactly what we had before, nor what we have now.

I'll update both still.

PR #364, opened 2020-12-03
Add reference article about events

bladealslayer:


src/docs/references/events/index.md
```diff
@@ -0,0 +1,392 @@
...
+    ...
+  }
+}
+```
+
+### Event sources
+
+The following table lists all possible event sources:
+
+Source | Description
```
ovan, line 371: Interestingly I've seen tegel as source a few times in our real event data when testing out the CLI. I'm not sure right now what is causing that and if those events should be there in the first place.

bladealslayer: Good catch, there are indeed almost 40k such events, all of which have been resolved as `:listing/updated`. I checked a few and they don't have `previousValues`. Looking at the code, I think that's because we detect `:listing/bookings` updates and don't ignore that attribute. That's a bug. I'll make PR to fix it and also delete those events from the db. @ovan

src/docs/references/events/index.md
```diff
@@ -7,26 +7,44 @@ ingress: Reference documentation for Flex marketplace events.
...
+Flex does not retain event data forever. Flex maintains a history of
+all marketplace events for 90 days in production marketplaces and for
+7 days in test and demo marketplaces.
 
 ## Event data
 
-The event object contains data about the event itself, as well as about the Flex
-Integration API resouce for which the event is about (`listing`, `user`,
+The event object contains data about the event itself, as well as
+about the Flex resource which the event is about (`listing`, `user`,
```
bladealslayer, line 47: Should we still emphasize more somewhere that events follow the Integration API resources (and not, say, the Marketplace API resources)? That was my intention here and where using the `INTEGRATION_API_RESOURCE_TYPE/EVENT_SUBTYPE` notation. Or do you see that as inaccurate?

ovan: @bladealslayer If you remember the "data models of Flex" diagram we sketched, the idea there was that Integration API exposes a canonical resource model. So my intention here was to reduce the prominence of Integration API as the driver for the data model and rather say that this is our default data model, which is exposed in events and by the Integration API.

src/docs/references/events/index.md
```diff
@@ -0,0 +1,414 @@
...
+attribute of the event contains the full data for the corresponding
+API resource **after** the event. The `resource` object contains keys
+corresponding to the attributes that the Integration API resource has,
+including it's `id`. For list of those attributes, see the Integration
+API reference for each resource type (e.g.
+[user](https://www.sharetribe.com/api-reference/integration.html#users),
+[listing](https://www.sharetribe.com/api-reference/integration.html#users),
+etc).
+
+In addition, when the resource has one or more 1-to-1 [TODO relationships]() with
```
ovan, line 247: @bladealslayer I wasn't sure where you intended to link here so couldn't fix the missing link

src/docs/references/events/index.md
bladealslayer: I guessed right that this was the reason then :) I guess my worry is that from reader's perspective, they don't yet have much to point them in that direction. It seemed to me easiest to give the Integration API as a reference point, as this is the only canonical data model they see. Maybe we'll gradually build that picture from now on with events, webhooks, etc.

src/docs/references/events/index.md
bladealslayer: Just the generic explanation for relationships: https://www.sharetribe.com/api-reference/#including-related-resources

src/docs/references/events/index.md
bladealslayer: Addressed in https://github.com/sharetribe/sharetribe-custom/pull/1974

PR #369, opened 2020-12-14
Events refactored structure

Gnito:
Suggestion to different structure inside this article. This way event formats are not discussed before they are defined.
Screenshot 2020-12-14 at 20 08 24

So, main suggestion:
- Removing format discussion from `## Event data`
- Swapping the order of `### Event data formats` & `### Resource data and previous values`
- Moving the preassumptions of `### Resource data and previous values` to the beginning of the example
- Opening format examples a bit: first `...` > `// other event keys: "eventType", "sequenceId", "createdAt", "marketplaceId", "source", "resourceId", "resourceType", "auditData"`
- `### Event sources` > `## Event sources`

https://deploy-preview-369--sharetribe-flex-docs-site.netlify.app/docs/references/events/

### DIFF
--- a/src/docs/references/events/index.md
+++ b/src/docs/references/events/index.md
@@ -48,43 +48,7 @@ about the Flex resource which the event is about (`listing`, `user`,
 `message`, etc), including how the resource changed.
 
 The exact shape of the event data differ, depending on which means it is
-accessed with, but the information contained is the same, regardless. For
-instance, when accessed via the Integration API, the event and associated
-resource data is formatted in the same way all Integration API resources are
-(with `id`, `type` top-level attributes and with `relationships` object
-containing the IDs of related resources). On the other hand, if the event is
-viewed via Flex CLI, the event and resource data is not normalized in that way
-and is inlined in a simplified form. We will refer to this as *native* event
-data format below. For details on the difference, see [Native and Integration
-API event data format](#native-and-integration-api-data-format) below.
-
-Here is an abbreviated example of event data in the native format, including all
-top-level event attributes. For a full example, including `resource` and
-`previousValues`, see [Resource data and previous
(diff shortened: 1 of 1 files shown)


src/docs/references/events/index.md
```diff
@@ -130,10 +96,90 @@ that have the exact same `createdAt` timestamp, so applications should
 not use the timestamp as a strict criteria to determining event
 ordering or to decide which events are processed and which are not.
 
+### Event data formats
+
+There are two data formats for events: **native** and **Integration API** event data formats.
+
+If the event is viewed via Flex CLI, the event and resource data is not normalized. Instead, it is inlined in a simplified form. We will refer to this as *native* event data format. On the other hand, when accessed via the Integration API, the event and associated resource data is formatted in the same way all Integration API resources are.
```
bladealslayer, line 103: @Gnito do you think it's ok to reverse the order in which formats are introduced here or somehow change then the description of the native format? The way it's presented now, it is defined in contrast with the Integration API format (i.e. "not normalized", "simplified"), which doesn't make sense as first thing before the Integration API format is explained. What was the issue you saw with the original sentence:

> when accessed via the Integration API, the event and associated
resource data is formatted in the same way all Integration API resources are
(with `id`, `type` top-level attributes and with `relationships` object
containing the IDs of related resources). On the other hand, if the event is
viewed via Flex CLI, the event and resource data is not normalized in that way
and is inlined in a simplified form.

Gnito: @bladealslayer No opinion about this. I just tried to combine different sentences when swapped the article structure. 
My suggestion is that if you agree with the structural change, you just merge this PR to yours and then rewrite the actual content under the headings if needed. :)

PR #384, opened 2021-01-11
Update OIDC proxy doc

OtterleyW:
Add information about testing in Heroku because localhost won't work

### DIFF
--- a/src/docs/cookbook-social-logins-and-sso/setup-oidc-proxy/index.md
+++ b/src/docs/cookbook-social-logins-and-sso/setup-oidc-proxy/index.md
@@ -1,7 +1,7 @@
 ---
 title: How to set up OpenID Connect proxy in FTW
 slug: setup-open-id-connect-proxy
-updated: 2020-12-16
+updated: 2021-01-11
 category: cookbook-social-logins-and-sso
 ingress:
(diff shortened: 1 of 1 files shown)


src/docs/cookbook-social-logins-and-sso/setup-oidc-proxy/index.md
```diff
@@ -18,6 +18,15 @@ into an OpenID Connect ID token that can be used to validate user login
 in Flex. With this approach, FTW will serve as an identity provider
 towards Flex.
 
+Flex verifies the ID token by fetching the JSON Web Key that is hosted
+by your FTW server and using that to unsign the token. A consequence of
+this is, that the JSON Web Key needs to be publicly available. This
+means that the proxy setup will not work in localhost. To test out the
+LinkedIn login, you should e.g.
+[deploy your FTW changes to Heroku](/tutorial-branding/deploy-to-heroku/)
+so that the .well-known endpoints are publicly reachable. You can read
```
lyyder, line 27: How about "so that the `.well-known/*` endpoints are publicly available", and then drop the "You can read more about"?

lyyder (commented):
👍 , the `updated` attribute can also be updated.

PR #399, opened 2021-02-19
Add overview of different payment methods

sktoiva:
The three articles changed:

- Added: Payment methods background article
- How payment intents work article
- Enable PaymentIntents cookbook article


src/docs/background/payment-methods-overview/index.md
```diff
@@ -0,0 +1,121 @@
...
+process and include new transitions. These should initiate the push
+payment and potentially skip the preauthorization phase. When you
+request a new push payment, you need to use the new action
+[`stripe-create-payment-intent-push`](../references/transaction-process-actions/#actionstripe-create-payment-intent-push).
+The example below describes the minimum recommended changes in the two
+transitions: `request-payment-push` and `confirm-payment-intent`.
+
+![Push payment process](push-payment-process.png 'Push payment process example.')
+
+## Further reading
```
bladealslayer, line 113: I think we could also add a short section explaining that when implementing the new payment methods, the Stripe's own instructions on how to deal with them with payment intents apply and the parts where something is done from server (basically just the payment intent creation), that's covered by Flex and remind that Flex exposes the payment intent ID and client secret through the transaction protected data to the customer.

src/docs/background/payment-methods-overview/index.md
```diff
@@ -0,0 +1,121 @@
...
+
+If you wish to enable push payments, you need to change your transaction
+process and include new transitions. These should initiate the push
+payment and potentially skip the preauthorization phase. When you
+request a new push payment, you need to use the new action
+[`stripe-create-payment-intent-push`](../references/transaction-process-actions/#actionstripe-create-payment-intent-push).
+The example below describes the minimum recommended changes in the two
+transitions: `request-payment-push` and `confirm-payment-intent`.
+
+![Push payment process](push-payment-process.png 'Push payment process example.')
```
bladealslayer, line 111: Minor thing - the "tx process continues" shape could be removed and instead the arrow can point into nothing. I think we've used that kind of notation elsewhere (e.g. here: https://www.sharetribe.com/docs/background/off-session-payments-in-transaction-process/#gatsby-focus-wrapper)

bladealslayer (commented):
Looks great to me so far! A couple of suggestions and one more fundamental thing I just noticed:

Stripe seems to have used push/synchronous terminology earlier with Sources, but actually the newer docs don't seem to use that at all. Instead they talk about "[immediate notification](https://stripe.com/docs/payments/payment-methods#payment-notification)" and "delayed notification". I didn't exactly spot how they refer to push/pull and I think they don't really mention that classification everywhere. The closest I found is "business initiated confirmation" and "customer-initiated confirmation" or "customer-authenticated". I think they nowadays prefer to group the payment methods, instead, as bank transfers, bank redirects, wallets, etc. It kinda makes sense, but also blurs the lines, because that doesn't align to the push/pull thing.

Should we change our wording and talk about "immediate notification" instead of "synchronous" and "customer-initiated" (perhaps), instead of "push"? Or mention that Stripe docs use both? Something like:

> Stripe describes as synchronous push payment methods

into

> Stripe describes as customer-initiated payments with immediate notification (or as synchronous push payment methods in other docs)

src/docs/background/payment-methods-overview/index.md
```diff
@@ -0,0 +1,158 @@
...
+
+If you wish to enable push payments, you need to adapt your transaction
+process. For instance, you need to add a new transition that includes
+the
+[stripe-create-payment-intent-push](/references/transaction-process-actions/#actionstripe-create-payment-intent-push)
+action. Further, because push payments do not have a preauthorization
+phase, it is recommended to avoid that in the transaction process and
+use an *instant booking* type of flow. The example below describes the
+minimum recommended changes in the two transitions:
+`request-payment-push` and `confirm-payment-intent`. The example
```
Gnito, line 121:
```suggestion
`request-push-payment` and `confirm-payment-instant-booking`. The example
```

src/docs/background/payment-methods-overview/index.md
```diff
@@ -0,0 +1,141 @@
...
+phase, it is recommended to avoid that in the transaction process and
+use an _instant booking_ type of flow. The example below describes the
+minimum recommended changes in the two transitions:
+`request-push-payment` and `confirm-payment-instant-booking`. The example
+illustrates how you can still use the preauthorization step for card
+payments.
+
+![Push payment process](push-payment-process.png 'Push payment process example.')
+
+### Handling payment methods in the UI
```
Gnito, line 127:
```suggestion
### Handling payment methods in the client app
```

part of this happens in client app's server, part in UI layer

src/docs/references/transaction-process-actions/index.md
```diff
@@ -915,12 +915,32 @@ requires that the transition is made from a trusted context.
...
 
+After the PaymentIntent is created, it's ID and client secret are
+temporarily accessible in the Flex transaction protected data under the
+`stripePaymentIntents` key in the following form:
+
+```json
+{
+  "default": {
+    "stripePaymentIntentId": "pi_1EXSEzLSea1GQQ9x5PnNTeuS",
+    "stripePaymentIntentClientSecret": "pi_1EXSEzLSea1GQQ9x5PnNTeuS_secret_xxxxxxxxxxxxxxxxxxxxxxxxx"
```
Gnito, line 934: Btw. does it matter if these are obfuscated or not?
Here's one shown as it is:
https://deploy-preview-399--sharetribe-flex-docs-site.netlify.app/docs/background/payment-intents/#step-2-collect-payment-information-and-handle-customer-actions

src/docs/background/payment-methods-overview/index.md
bladealslayer: @Gnito which part is happening in the server?

src/docs/references/transaction-process-actions/index.md
bladealslayer: It doesn't really matter. Do you think obfuscating that has any downside? The format of that value anyway is not controlled by us and can change, so people shouldn't rely on what exactly it looks like.

PR #409, opened 2021-03-03
SEO updates

kpuputti:


src/docs/cookbook-payments/customize-pricing/index.md
```diff
@@ -1,6 +1,6 @@
 ---
-title: Customize pricing
-slug: customize-pricing
+title: How to customize pricing
+slug: how-to-customize-pricing
```
Gnito, line 3: I don't have a real preference here, but why to add this extra stuff to URL? 
I think `/cookbook-payments/customize-pricing` is much more to the point (less noise) and then there isn't any need for redirections.

kpuputti: The original task was to separate the articles titles where we had two articles with title "Customize pricing". I just thought that it would be better (SEO?) if the URL slug matches the title.

Gnito: My guess is that there is no effect - and anyway, it's pretty much black magic to start thinking if generic string like "how-to" in URL has positive or negative effect on anything.

That being said, if we have 2 articles with same name or slug, then separating those makes sense.

PR #443, opened 2021-06-03
Test YouTube iframe video embed

thomasmalbaux:
Just a test to make sure this works elsewhere than locally


thomasmalbaux:
It works, I just wanted to test. No further work needed.

PR #477, opened 2021-09-30
Start updating off-session payments documentation

SariSaar:


src/docs/background/off-session-payments/index.md
```diff
@@ -34,7 +34,10 @@ booking period.
 Suppose your sauna rentals marketplace should allow customers to book
 saunas up to a year in advance, but the customer is charged only a
 certain time in advance before the booking. The figure below illustrates
-how a part of your transaction process might look like.
+how a part of your transaction process might look like. The
+[flex-example-processes Github repository](https://github.com/sharetribe/flex-example-processes)
```
SariSaar, line 38: TODO: Once the example process is in the repo, this link could lead straight to the correct example process

src/docs/background/off-session-payments/index.md
```diff
@@ -62,7 +65,7 @@ kicks in), the payment card might have expired, etc. It is therefore
 always important to allow for a fall-back payment path in your
 transaction process. Since only one transition from a state can be
 triggered automatically, this fall-back payment path must be defined to
-trigger upon a user action.
+trigger upon a user action, as in the example.
```
SariSaar, line 69: TODO: Add a 'drawer' component to explain the Stripe off-session dynamic with :use-customer-default-payment-method and link to the article. It's mentioned in the image, but linking from the image is not possible

SariSaar: Added

src/docs/background/off-session-payments/index.md
```diff
@@ -62,7 +83,7 @@ kicks in), the payment card might have expired, etc. It is therefore
...
 
 You can build upon this example and extend it to make the payment
 process more robust. For instance, in case the customer fails to pay for
```
SariSaar, line 89: Feedback welcome: Would it make sense to have a section on 'what to pay attention to if you implement a process like this in FTW'? I can make those additions as well - not necessarily on a tutorial-specific level but as a general 'good to be aware' listicle.

shareoc: I think a listicle sounds like a good idea! The more information is available here, the less likely it is that people will target related questions at support.

SariSaar: First attempt at listicle written, appreciate any feedback!

src/docs/background/off-session-payments/index.md
```diff
@@ -22,7 +22,7 @@ pattern.
 Flex API has capabilities for
 [saving payment card details for future use](https://www.sharetribe.com/api-reference/marketplace.html#stripe-customer).
 In addition, it is possible to configure your transaction process in
-such a way, that the customer is charged automatically off-session at
+such a way that the customer is charged automatically off-session at
```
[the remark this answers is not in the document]
SariSaar: Yes :) Fixed!

src/docs/background/off-session-payments/index.md
```diff
@@ -72,3 +96,56 @@ provider to post a review of the customer. In addition, you may consider
...
+
+### Handling delayed manual payment
+
+If the automatic payment succeeds, the customer does not need to take
+further action on the transaction before the review process. Manual
+payment, on the other hand, does require a new user flow in the FTW
+template. CheckoutPage.js is set up to
+[handle payments toward Stripe](https://github.com/sharetribe/ftw-daily/blob/6e856692d0b7f3a7a1110894967a229a70ae0ce0/src/containers/CheckoutPage/CheckoutPage.js#L795),
+so the simplest option is that once the customer has confirmed the
+booking and triggered the transition to create a payment intent, they
```
lyyder, line 141: Should this say that the provider has confirmed the booking? Could it be mentioned here something like _once a payment has not automatically gone through and the customer has manually triggered the transition to create a payment intent..._ or something like that? Would that make sense here?

lyyder (commented):
Nice work 👏

The list for things to consider in FTW side is super. With the process image, one idea would be to increase the size a bit, or somehow make more room for the transitions so that it's clear that which transition each text is related to. So the transitions could be longer or the descriptions could be placed somehow so that they fit between the states, making it more explicit that they are about the transition.

src/docs/background/off-session-payments/index.md
```diff
@@ -72,3 +96,56 @@ provider to post a review of the customer. In addition, you may consider
...
+contains information about the transaction's last transition and its
+timestamp.
+
+### Separating order from payment
+
+In the default transaction process and default FTW flow, the order is
+initiated and processed on
+[CheckoutPage.js](https://github.com/sharetribe/ftw-daily/blob/6e856692d0b7f3a7a1110894967a229a70ae0ce0/src/containers/CheckoutPage/CheckoutPage.js#L355)
+using
+[initial values from ListingPage.js](https://github.com/sharetribe/ftw-daily/blob/6e856692d0b7f3a7a1110894967a229a70ae0ce0/src/containers/ListingPage/ListingPage.js#L122).
```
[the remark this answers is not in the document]
SariSaar: Good considerations. I set the Github links to direct to master on component level and fixed the wording so there's some guidelines re: where to look more specifically, as especially CheckoutPage.js is fairly massive.

src/docs/background/off-session-payments/index.md
```diff
@@ -72,3 +96,56 @@ provider to post a review of the customer. In addition, you may consider
...
+Since the off-session payment process separates initiating the order
+(i.e. creating a booking, setting line items in a privileged transition)
+from payment (creating and further processing Stripe payment intent), it
+is important to pay attention to the way you want to handle that
+separation.
+
+- What happens when user clicks 'Request to book' on ListingPage.js?
+- Where is the API call made to create a booking and set line items?
```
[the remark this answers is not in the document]
SariSaar: Fixed!

src/docs/background/off-session-payments/index.md
SariSaar: The original thought was 'once the customer has confirmed the _payment_', but the more explicit wording was better, so fixed in that style!

SariSaar:
There was a ticket asking about implementing a 'pre-order' functionality, and we could add a comment here that a similar logic could be implemented for a product marketplace that wants to support pre-orders.

SariSaar:
Yes, that was the logic I was thinking for a pre-order.

PR #483, opened 2021-10-05
fix date typo and adds 1 example

Luis-RG:
added 1 example cause I got super confused by "example 2". Probably not needed for developers that are super familiar with timestamps, but well, it can't hurt.

### DIFF
--- a/src/docs/references/availability/index.md
+++ b/src/docs/references/availability/index.md
@@ -1,7 +1,7 @@
 ---
 title: Listing availability management
 slug: availability
-updated: 2020-09-24
+updated: 2020-10-05
 category: references
 ingress: Reference documentation for listing availability management.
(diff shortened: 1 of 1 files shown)


src/docs/references/availability/index.md
```diff
@@ -84,9 +84,15 @@ An exception with start `2018-11-26T00:30.000+01:00` and end
 
 ###### **Example 3:**
 
-Given exception 1 from `2017-11-26T10:00:00.000Z` to
+An exception with start `2018-11-26T00:30.000+01:00` and end
```
shareoc, line 87: 2018-11-26T00:30.000+01:00 is invalid timestamp, should be 2018-11-26T00:30:00.000+01:00

PR #515, opened 2021-12-28
Add note of one automatic transition executed per state

SariSaar:
A transaction process state can have several automatic transitions scheduled, however only one automatic transition can be executed from a single state. This is currently not mentioned in Flex Docs. Add a mention of this limitation to the documentation.

### DIFF
--- a/src/docs/references/transaction-process-format/index.md
+++ b/src/docs/references/transaction-process-format/index.md
@@ -344,6 +344,13 @@ in the
 [Transaction process time expressions](/references/transaction-process-time-expressions/)
 reference article.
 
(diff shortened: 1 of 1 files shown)


SariSaar:
Need to fact check this from a Flex dev, so in drafts until that happens :)

lyyder (approved):
LGTM!

One thing that could be also mentioned, is that if the transition that is first scheduled fails, no transition that would be scheduled later is executed. In the off-session payments flow this is handles so that there's a scheduled message that is sent after a scheduled transition. If the transition fails, then the message will notify a transaction party to interact with the transaction.

PR #516, opened 2021-12-29
Mention problematic Stripe countries in setup cookbook

SariSaar:
FTW does not support all countries where Stripe Custom Connect accounts are available. This PR adds information on the reasons why Brazil, India and Hungary aren't supported, so users can determine whether adding them through custom development is worth the effort for them.

### DIFF
--- a/src/docs/cookbook-payments/set-up-and-use-stripe/index.md
+++ b/src/docs/cookbook-payments/set-up-and-use-stripe/index.md
@@ -1,7 +1,7 @@
 ---
 title: Set up and use Stripe
 slug: set-up-and-use-stripe
-updated: 2019-03-27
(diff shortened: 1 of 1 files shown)


shareoc (approved):
Everything looks good!

src/docs/cookbook-payments/set-up-and-use-stripe/index.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: Set up and use Stripe
 slug: set-up-and-use-stripe
-updated: 2019-03-27
+updated: 2021-12-29
```
SariSaar, line 4:
```suggestion
updated: 2022-01-04
```

PR #518, opened 2022-01-01
Fix a broken link

thomasmalbaux:


thomasmalbaux:
Ping @SariSaar, I've noticed that the broken link was introduced at https://github.com/sharetribe/flex-docs/commit/3b66e5e161c4fdaa460d1ea368f2b9ec17e02570. Just a ping so that you know that `/docs` has to be part of the link target :-)

PR #519, opened 2022-01-11
Cookiebot

kpuputti:


Gnito (approved):
LGTM. (I didn't test fallback)

Gnito:
It's probably worth adding some comment to the `docs` directory about how this should be used:
https://github.com/sharetribe/flex-docs/tree/master/docs

PR #529, opened 2022-01-21
Add new cookbook article

shareoc:
Includes a new Cookbook article on how to create a like-feature using events. Full code is available at: https://github.com/shareoc/events-and-likes/tree/likes-demo


src/docs/cookbook-events/like-feature/index.md
```diff
@@ -0,0 +1,527 @@
...
++       setTimeout(() => {pollLoop(lastSequenceId);}, delay);
++     })
++   });
+};
+```
+
+You can now run the script locally following the instructions [here](https://github.com/sharetribe/flex-integration-api-examples#getting-started).
+
+## Increment the counter in the UI
+You can now click on the likes button and the amount of likes will be updated to listing extended data. However, the like counter doesn't get updated immediately due to the slight latency in our events listener. We can add a increment the counter using React state. In addition, we'll add conditional styling to reflect has the listing been liked or not.
```
lyyder, line 464: I think this could elaborate a bit more on the idea that the likes will be eventually consistent and that for UI reasons will update the like count also in the front end so that the user will know that the listing is liked but the actual change happens via events. I know it's already mentioned there but I think it could be explained in a bit more detail so that readers who are not so familiar with this kind of "eventually consistent" data flow will understand the idea of this artificial state change in the UI.

lyyder: And discuss that this is not totally necessary but what would be the downsides of not doing this UI state change. And on the other hand, what problems may this introduce if the poll delay in the events listener is long and it takes a long time for the likes data to update in the API.

lyyder (commented):
Really good work. The text is easy to follow and the UI changes described here look nice.

Some ideas on things that could also be discussed in this article. However, these topics are not only related to this feature but all worker/batch type of tasks so these are not necessarily required in this article:

- Now the events listener stores that last processed event sequence number in a file, right? This article could mention what other alternatives are there (some DB maybe) and what are the treadeoffs: a DB is harder to set up, can disk/file access be assured in server env if file is used and so on
- How to run this in production? What kind of setup could be used, mention that this does not scale horizontally and should be run on one instance only

I'd also consider writing this as it's own example in the integration API repo. It would make this article shorter. The example in the repo would need to mention that it's supposed to be used together with this cookbook but it could serve a purpose on it's own too. What do you think? Any downsides to this approach?

lyyder:
> I think it would be good to have a separate repository with the example code files from tutorials/guides. It can be easier for some to just get instant access to the code than following the tutorial.

@shareoc good idea. Having the code changes in the articles has it's downsides. A distinct repo on the other hand would need to be kept up to date with the latest FTW development. Maybe this is a topic that we could discuss in another forum.

src/docs/cookbook-events/like-feature/index.md
```diff
@@ -0,0 +1,527 @@
...
+```shell
+└── src
+    └── containers
+        └── ListingPage
+            ├── SectionLikes.js
+            ├── ListingPage.js
+            └── ListingPage.module.css
+```
+
+We'll create a new component that renders an svg and the amount of likes:
```
SariSaar, line 33: This might be a matter of preference, but for me it's easier to read a guide like this if there's some explanation between the heading and the code example, so there's some context as to what I should be paying attention to.

src/docs/cookbook-events/like-feature/index.md
```diff
@@ -0,0 +1,527 @@
...
+You might be wondering why we don't we just store the 'likes' directly to listing extended data. There are two reasons for that: saving likes in user extended data tells us what listings a user has liked. In addition, we avoid a race condition caused by two users liking a listing simultaneously.
+
+A downside to using events is that the like count may take a moment to update. In this guide, we temporarily increment the like value in the UI to hide the delay. 
+
+You can download the files used in this guide from our example repository here, or follow along using the embedded code snippets.
+
+## Create a UI component the user can interact with
+We start by creating an icon that users can interact with to like or dislike a listing. Next to the icon, we'll display the number of likes. For the UI component, we'll create a new subcomponent `SectionLikes.js` in the ListingPage directory. 
+
+###### Step 1: Create a new file
```
SariSaar, line 22: Is there a reason why these use H5 instead of H3 that would be the next step down?

src/docs/cookbook-events/like-feature/index.md
```diff
@@ -0,0 +1,527 @@
...
++   if (!updateLikesInProgress && currentUser) {
++     onUpdateLikes(listingId, currentLikes);
++    }
++  }}>
+```
+
+Now, clicking on the icon will update either add or remove the listing to the user exteneded data with the key `likedListings`. However, as liked listings are only saved into user extended data at the moment, the number of likes still remains at zero. 
+
+## Listening to events
+Using events we can react to changes in users' extended data. To do this we'll listen to the /events/query endpoint, filter out relevant events and finally update listing extended data. We'll use the [`notify-new-listings.js`](https://github.com/sharetribe/flex-integration-api-examples/blob/master/scripts/notify-new-listings.js) script in the integration-api-examples as a basis for our new script. Make sure to follow the instructions at the [root of the repository](https://github.com/sharetribe/flex-integration-api-examples#getting-started) if you're unsure how to run the script locally.
```
SariSaar, line 359: I'd suggest adding a link on the `/events/query endpoint` text pointing to the API ref

src/docs/cookbook-events/like-feature/index.md
```diff
@@ -0,0 +1,527 @@
...
+updated: 2021-01-20
+category: cookbook-events
+ingress:
+  Following this guide, you'll be able to build a feature that allows users to like listings using events and user extended data
+published: true
+---
+
+Events represent changes in marketplace data resources. In this guide, we use events to listen to changes in user extended data. We react to these events by updating a 'likes' value in listing extended data. In addition, we display the number of likes on the listing page and allow users to interact by clicking on a 'like' -button. 
+
+You might be wondering why we don't we just store the 'likes' directly to listing extended data. There are two reasons for that: saving likes in user extended data tells us what listings a user has liked. In addition, we avoid a race condition caused by two users liking a listing simultaneously.
```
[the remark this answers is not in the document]
shareoc: Good idea. I wrote a longer section on the architectural decisions.

src/docs/cookbook-events/like-feature/index.md
```diff
@@ -0,0 +1,527 @@
...
+const SectionLikes = props => {
+  const { 
+    publicData, 
+  } = props;
+
+  const likes = publicData?.likes ? publicData.likes : 0;
+
+  return (
+    <span className={css.heartIcon}> 
+      <IconHeart /> <FormattedMessage id="ListingPage.likes" values={{ likes: likes }} />
```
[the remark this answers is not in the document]
shareoc: Yup, that approach works and is a lot simpler.

src/docs/cookbook-events/like-feature/index.md
```diff
@@ -0,0 +1,527 @@
...
+
+Now, you should be able to see a heart-shaped icon with a like counter next to it if you navigate to the listing page:
+
+![Example of heart icon](heart-icon.png 'Example of heart icon')
+
+
+## Update user extended data
+To update the user extended data, we'll need to make some changes to ListingPage.duck.js. We'll need to import the currentUserShowSuccess function from `user.duck.js` to update the current user. In addition, we'll be adding a new action type, action creator and reducer to the file. For more information on how Redux is setup in FTW, refer to our [article on Redux]('/ftw-data-flow/ftw-redux').
+
+###### Step 1: Import currentUserShowSuccess
```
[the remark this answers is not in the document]
shareoc: All headings up to H4 will appear in the content table. H5 is stylised as a heading whereas H6 is stylistically more appropriate here. However, from the perspective of semantic HTML, this isn't the best approach as you should avoid skipping heading levels.

src/docs/cookbook-events/like-feature/index.md
```diff
@@ -0,0 +1,527 @@
...
+  );
+};
+
+```
+
+Next, we'll add a function that updates the 'likes' value in listing extended data by calling the Integration API listings/update endpoint.
+
+###### Step 2: Add a function that updates the 'likes' value:
+```js
+  const updateListing = (listingId, likeCount) => {
```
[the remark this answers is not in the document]
shareoc: I couldn't come up with a hypernym for addition and subtraction. I added a function doc string and renamed it to likeAddition

src/docs/cookbook-events/like-feature/index.md
shareoc: See this comment: https://github.com/sharetribe/flex-docs/pull/529#discussion_r803502726

src/docs/cookbook-events/like-feature/index.md
```diff
@@ -0,0 +1,527 @@
...
+  error: true,
+})
+
+```
+
+###### Step 6: Add new thunk
+```jsx
+export const updateLikes = (listingId, currentLikes) => (dispatch, getState, sdk) => {
+  dispatch(updateLikesRequest());
+  dispatch(fetchCurrentUser());
```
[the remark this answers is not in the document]
shareoc: Hmm, I'm actually not sure about this one. Maybe @Gnito would know?

shareoc:
> Really good work. The text is easy to follow and the UI changes described here look nice.
> 
> Some ideas on things that could also be discussed in this article. However, these topics are not only related to this feature but all worker/batch type of tasks so these are not necessarily required in this article:
> 
> * Now the events listener stores that last processed event sequence number in a file, right? This article could mention what other alternatives are there (some DB maybe) and what are the treadeoffs: a DB is harder to set up, can disk/file access be assured in server env if file is used and so on
> * How to run this in production? What kind of setup could be used, mention that this does not scale horizontally and should be run on one instance only
> 
> I'd also consider writing this as it's own example in the integration API repo. It would make this article shorter. The example in the repo would need to mention that it's supposed to be used together with this cookbook but it could serve a purpose on it's own too. What do you think? Any downsides to this approach?

I think these pointers could be addressed in their own articles. I'll make a pull request to the Integration API but as the whole article is structured as 'follow and code along', maybe it's better to keep the script embedded in this article?

src/docs/cookbook-events/like-feature/index.md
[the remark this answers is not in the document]
lyyder: Could skipping heading levels have an effect on SEO? Would bolded texts work if we don't want to skip headings? Using heading-tags is probably better for screen readers. Maybe it's fine just like this.

src/docs/cookbook-events/like-feature/index.md
```diff
@@ -0,0 +1,536 @@
...
+  if (currentLikes === null) return previousLikes;
+  else return previousLikes.length < currentLikes.length ? getDifference(currentLikes, previousLikes) : getDifference(previousLikes, currentLikes);
+}
+
+const getLikeCount = (previousLikes, currentLikes) => {
+  return previousLikes === null || previousLikes.length < currentLikes.length ? 1 : -1;
+}
+
+// Reducer returns an object with listing ID's as keys and amount of likes as values
+const sortEvents = events => {
```
lyyder, line 424: Instead of `sort` in the fn name, the action here could be `group`, as the likes are grouped for each listing? So `groupEvents` or even `groupLikes`.

src/docs/cookbook-events/like-feature/index.md
```diff
@@ -0,0 +1,536 @@
...
+
+To allow users to like listings, we introduce a UI-element users' can interact with to like a listing. However, when we update a listing's extended data as a reaction to user input, we are prone to a race condition. Fortunately, Flex provides a powerful feature called [events](references/events/) that we can use to solve the problem. We can listen to events using a script that polls the [events endpoint](https://www.sharetribe.com/api-reference/integration.html#query-events). As events are handled sequentially, the script can update the listing's extended data while avoiding race conditions. A downside to using events is that the like count may take a moment to update. We can mask this delay by temporarily incrementintg the like value in the UI.
```
lyyder, line 19: s/users'/users ?

src/docs/cookbook-events/like-feature/index.md
shareoc: Added some explanations below the H5 headings!

src/docs/cookbook-events/like-feature/index.md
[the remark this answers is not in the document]
shareoc: Something like this?
```
return dispatch(fetchCurrentUser()).then(response => {
     return sdk.currentUser
     .updateProfile({ privateData: { likedListings } }, queryParams)
     .then(() => {
        dispatch(updateLikesSuccess(response));
        const entities = denormalisedResponseEntities(response);
        if (entities.length !== 1) {
          throw new Error('Expected a resource in the sdk.currentUser.updateProfile response');
        }
        const currentUser = entities[0];
        // Update current user in state.user.currentUser through user.duck.js
        dispatch(currentUserShowSuccess(currentUser));
      })
      .catch(e => {
        dispatch(updateLikesError(storableError(e)))
      });
  });
};
```

PR #532, opened 2022-02-04
Add first draft of migration docs improvements

SariSaar:
Improve operator guide documentation on the Intermediary format
- Add code examples to illustrate the text throughout
- Add mention that user performs the extract and transform steps of migration, and Sharetribe loads the data into Flex
- Expand explanations of referencing and aliases in Intermediary
- Add detailed validation schemas with `required` / `optional` information for each resource available for migration
- Add explanations of value types used in Intermediary
- Add mentions of data sharing methods as well as using a representative subset of the data for test migrations

### DIFF
--- a/src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
+++ b/src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
@@ -15,16 +15,13 @@ case and you can find the outline for this process
 
 If you are running a marketplace outside the Sharetribe ecosystem, and
 would like to import data to Flex, it is possible. You will need to
-extract your data from its existing storage and transform it into a data
-format called Intermediary. Sharetribe will then validate the
-transformed data and load it into Flex.
+extract your data from its existing storage and transform it into
+Sharetribe's proprietary data format called Intermediary. Sharetribe
+will then validate the transformed data and load it into Flex.
 
 This article will outline how you can encode data into Intermediary
 format, give you syntax examples of the format, and provide instructions
 on the migration process.
-
-The tooling for these migrations is still in it's infancy and will
-require some technical knowledge and a lots of patience.
 
 ## Intermediary data
 
@@ -42,18 +39,19 @@ Intermediary is specified as
 following keys:
 
 - :ident - identifies the target marketplace by marketplace ID,
-- :data - a seq of marketplace data rows. Each row is a 2-tuple of id
-  and content.
+- :data - a [seq](https://clojure.org/reference/sequences) of
+  marketplace data rows. Each row is a 2-tuple (an array with two
+  elements) of id and content.
 
 You can find your marketplace ID in Flex Console > Build > General. Note
 that the anonymised test file needs to specify your test environment
 marketplace ID and your production data file needs to specify your
 production environment marketplace ID.
 
-The id part of row 2-tuple is specified as a tuple of 1 to 3 elements.
-The first element is always an id attribute and identifies the row type.
-The second element can be an alias or an import id. A 3 element version
-has id attr, import id and an alias.
+The id part of the data row 2-tuple is specified as a tuple of 1 to 3
+elements. The first element is always an id attribute and identifies the
+row type. The second element can be an alias or an import id. A 3
+element version has id attr, import id and an alias.
 
 ```
 ;; 1 element tuple
@@ -76,19 +74,21 @@ written by hand. Import ids are useful for programmatically generated /
 exported data.
 
 Note that import ids are only used in the import phase and they can not
-be mapped to existing resource ids. TODO: EXPLAIN BETTER! Content part
-of row 2-tuple is a map that is processed by tuple expansion function
-registered for the content type.
+be mapped to existing resource ids, because the final resource id is not
+created based on the import id. In other words, if e.g. your test
+marketplace already has data rows, you cannot reference those rows from
+the Intermediary file by the import ids. This means that all the data
+being referenced needs to be in the same Intermediary file.
 
-If you are not generating data by hand (TODO WHAT IS THE JUSTIFICATION
-HERE?), it is highly recommended that you use the 2 element form of the
-id tuple. This means that you don't use aliases but unique uuid's
-throughout the file to reference entities. You can read more about
-[generating UUIDs for your data here](https://www.uuidgenerator.net/api).
+If you are not generating data by hand, it is recommended that you use
+the 2 element form of the id tuple. This means that you use unique
+uuid's throughout the file to reference entities. You can read more
+about
+[generating UUIDs for your data here](https://www.uuidgenerator.net/dev-corner).
 
 #### Aliases
 
-If you currently use non-UUID ids in your data, you can use them to
+If you currently use non-UUID ids in your data, you can also use them to
 generate aliases for the data. Aliases must be unique across the file,
 and they cannot contain spaces.
 
@@ -109,11 +109,11 @@ return ':{type}/{type1stletter}{id}'
 When you already have data that is dependent on each other, such as
 listings related to users and reviews related to listings, you can
 reference the data within the import file. You can refer to resources by
-either alias or UUID with the `#im/ref` syntax. The supported references
-for each type, if any, are listed in the type description in the
-[Supported types](#supported-types) section below. Note that referencing
-data with an id or alias that does not exist in the same file causes a
-validation error.
+either alias or UUID with the `#im/ref` tagged element. The supported
+references for each type, if any, are listed in the type description in
+the [Supported types](#supported-types) section below. Note that
+referencing data with an id or alias that does not exist in the same
+file causes a validation error.
 
 ```
 ;; Referencing a listing by alias
@@ -165,22 +165,22 @@ Supported fields to migrate:
 [API reference listing resource](https://www.sharetribe.com/api-reference/marketplace.html#listing-resource-format)
 
 - Required attributes
-  - `:im.listing/createdAt`
-  - `:im.listing/title`
-  - `:im.listing/closed` OR `:im.listing/state`
+  - `:im.listing/createdAt` ([#inst](#timestamp))
+  - `:im.listing/title` (string)
+  - `:im.listing/closed` (boolean) OR `:im.listing/state`
   - accepted values for `:im.listing/state`
     - `:listing.state/published`
     - `:listing.state/closed`
     - `:listing.state/draft`
     - `:listing.state/pendingApproval`
 - Optional attributes - only include the key if it has a non-empty value
-  - `:im.listing/description`
-  - `:im.listing/location`
-  - `:im.listing/price`
-  - `:im.listing/author`
-  - `:im.listing/publicData`
-  - `:im.listing/privateData`
-  - `:im.listing/metadata`
+  - `:im.listing/description` (string)
+  - `:im.listing/location` ([#location](#location))
+  - `:im.listing/price` ([#money](#money))
+  - `:im.listing/author` (#im/ref)
+  - `:im.listing/publicData` (JSON)
+  - `:im.listing/privateData` (JSON)
+  - `:im.listing/metadata` (JSON)
 
 Supported references
 
@@ -193,7 +193,7 @@ Example syntax:
          #:im.listing{:createdAt #inst "2018-04-17T06:55:04.291-00:00"
                       :title "A solid rock sauna"
                       :description "A very nice solid rock sauna built solely of wood.\nHere's some more sensible stuff."
-                      :closed false
+                      :state :listing.state/pendingApproval
                       :location #im/location [23.12 21.21]
                       :price #im/money [12.12M "EUR"]
                       :publicData {:category "rock"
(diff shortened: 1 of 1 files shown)


src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
```diff
@@ -28,33 +32,102 @@ Intermediary is an edn (https://github.com/edn-format/edn) data format
...
-2-tuple is a map that is processed by tuple expansion function
+marketplace and have type [UUID](#uuid). Aliases can be used to
+reference rows from other rows. Aliases are useful when the data is
+written by hand. Import ids are useful for programmatically generated /
+exported data.
+
+Note that import ids are only used in the import phase and they can not
+be mapped to existing resource ids. TODO: EXPLAIN BETTER! Content part
+of row 2-tuple is a map that is processed by tuple expansion function
 registered for the content type.
```
SariSaar, line 81: This sounds like it's in Clojurian, so do we need an "in other words" in English as well or would there be e.g. a reference we could point to? This sentence doesn't really mean anything for a non-Clojure dev and I think most of the people reading this documentation will not speak Clojurian. :)

src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
```diff
@@ -28,33 +32,102 @@ Intermediary is an edn (https://github.com/edn-format/edn) data format
...
 
-If you are not generating data by hand, it is highly recommended that
-you use the 2 element form of the id tuple. This means that you don't
-use aliases but unique uuid's throughout the file to reference entities.
+If you are not generating data by hand (TODO WHAT IS THE JUSTIFICATION
```
SariSaar, line 83: What's the reasoning behind this recommendation?

src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
```diff
@@ -109,12 +215,202 @@ key like this:
...
+  #:im.image{:url "https://asset-url.someservice.com/path/to/img1.jpg"
+            :sortOrder 1
+            :listing #im/ref :listing/rock-sauna}]
+```
+
+#### Stripe account
+
+- Required attributes
+  - `:im.stripeAccount/stripeAccountId`
+  - `:im.stripeAccount/user` TODO: how about chargesEnabled and
```
SariSaar, line 291: Are `chargesEnabled` and `payoutsEnabled` optional or required? They're not references and they're not mentioned in the Intermediary spec

src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
```diff
@@ -109,12 +215,202 @@ key like this:
...
+
+```
+#:im.listing{...
+            :location #im/location [23.12 21.21]
+                      ...}
+```
+
+#### Money
+
+Money type is defined as [ amount, currency ]. _TODO What's the
```
SariSaar, line 378: Why is `#im/money` decimal and the Flex API endpoint type Money int?

src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
```diff
@@ -28,33 +32,102 @@ Intermediary is an edn (https://github.com/edn-format/edn) data format
 that allows defining data rows for marketplaces. It supports creating
 links between rows via references.
 
+We recommend that you use an .edn library to generate the data and
```
shareoc, line 35: Maybe we could provide a link to a library or two that could be used to generate and validate the syntax? Or are we aware of any we can recommend?

src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
```diff
@@ -28,33 +32,102 @@ Intermediary is an edn (https://github.com/edn-format/edn) data format
...
+validate the .edn syntax to ensure that the file is up to standard.
+
 ### Format
 
 Intermediary is specified as
 [a map](https://clojure.org/reference/data_structures#Maps) with the
 following keys:
 
-- :ident - identifies the target marketplace by marketplace ident,
+- :ident - identifies the target marketplace by marketplace ID,
```
shareoc, line 44: Like `map` on row 41, should `seq` and `2-tuple` link to a Clojure reference for developers not familiar with Clojure data structures?

shareoc:
Looks good, and I'm sure this will reduce friction for users migrating data from outside Sharetribe.

src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
SariSaar: That's a good idea! If @lyyder has ideas on what to recommend that would be great 👍

SariSaar:
@lyyder Since there's changes to the handling of stock in Intermediary in the stock migration project, do you think we should include it here or make an update later?

src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
lyyder: Good point. 2-tuple is somewhat a generic programming concept but `seq` is Clojure-specific. I think we can use the N-tuple wording in the document but when first mentioning it, we could write it out:
> 2-tuple (an array with 2 entries)

src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
lyyder: > Note that import ids are only used in the import phase and they can not
be mapped to existing resource ids.

This part is trying to say that intermediary can not be used to update existing resources, even though their ID would match an existing resource. Furthermore, what is defined as the ID of a resource in intermediary, will not be the ID of the resource in Flex. It is merely used in the import.

> Content part of row 2-tuple is a map that is processed by tuple expansion function registered for the content type.

This I suppose we can leave out. It's enough that we've explained that a resource consists of the ID tuple and content map, that's stated somewhere in the doc, right?

src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
lyyder: Good question. At least if data is programatically generated, it makes sense to use UUID generation to avoid ID collision instead of creating a custom alias based ID system.

src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
```diff
@@ -28,33 +32,102 @@ Intermediary is an edn (https://github.com/edn-format/edn) data format
...
 registered for the content type.
 
-If you are not generating data by hand, it is highly recommended that
-you use the 2 element form of the id tuple. This means that you don't
-use aliases but unique uuid's throughout the file to reference entities.
+If you are not generating data by hand (TODO WHAT IS THE JUSTIFICATION
+HERE?), it is highly recommended that you use the 2 element form of the
+id tuple. This means that you don't use aliases but unique uuid's
+throughout the file to reference entities. You can read more about
+[generating UUIDs for your data here](https://www.uuidgenerator.net/api).
```
lyyder, line 87: I suppose if intermediary is programmatically generated, we can assume that the environment/language used provides tooling to create UUIDs instead of integrating to a HTTP UUID API.

lyyder: So instead, we could refer to this on how to generate UUIDs in the programming language on has chosen: https://www.uuidgenerator.net/dev-corner

src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
```diff
@@ -28,33 +32,102 @@ Intermediary is an edn (https://github.com/edn-format/edn) data format
...
+// returns ':user/u123'
+return ':{type}/{type1stletter}{id}'
+```
+
+#### Referencing
+
+When you already have data that is dependent on each other, such as
+listings related to users and reviews related to listings, you can
+reference the data within the import file. You can refer to resources by
+either alias or UUID with the `#im/ref` syntax. The supported references
```
lyyder, line 112: In edn terminology, the `#im/ref` syntax is called [tagged element](https://github.com/edn-format/edn#tagged-elements).

lyyder: So here we could say "refer to resources ... with the `im/ref` tagged element".

src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
```diff
@@ -73,32 +146,65 @@ types:
...
+- [Profile images](#image)
+- [Listing images](#image)
+- [Stripe accounts](#stripe-account) - You need to use the same Stripe
+  keys in Flex as in your current service
+- [Reviews](#review)
+
+The migration process has some built-in validation for the types within
+the Intermediary file. Optional keys must have non-empty values, and if
+an optional key does not have a value, the key must be omitted for that
+resource. Empty strings are validated as empty values.
```
lyyder, line 160: 👍

src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
```diff
@@ -73,32 +146,65 @@ types:
...
+
+#### Listing
+
+Supported fields to migrate:
+[API reference listing resource](https://www.sharetribe.com/api-reference/marketplace.html#listing-resource-format)
+
+- Required attributes
```
lyyder, line 167: In the attributes list we could state the type of the data too. For example that `:im.listing/createdAt` is an [#inst](https://github.com/edn-format/edn#inst-rfc-3339-format) tagged element, title is a string, and so on.

lyyder (commented):
@SariSaar great work! Could not get to the end of the doc yet but there's a few comments.

src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
lyyder: IIRC, the Stripe integration in FTW fetches this data and it can be accessed that way.

src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
```diff
@@ -109,12 +215,202 @@ key like this:
...
+
+- [Location](#location)
+- [Money](#money)
+- [UUID](#uuid)
+- [Timestamp](#timestamp)
+
+#### Location
+
+Location is given as
+[LatLng type](https://docs.mapbox.com/android/maps/api/9.6.0/com/mapbox/mapboxsdk/geometry/LatLng.html)
```
lyyder, line 379: Would it be enough if we just stated that it's a 2-tuple of latitude and longitude?

src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
```diff
@@ -109,12 +215,202 @@ key like this:
...
+representing the latitude and longitude of the location. Both latitude
+and longitude are 32 bit floats. The `location` attribute is used in the
+FTW template to provide listing location coordinates through either
+Mapbox or Google maps.
+
+If your source data has addresses specified, and you would like to use
+them to determine locations, you can use e.g. the Mapbox
+[forwardGeoCode](https://github.com/sharetribe/ftw-daily/blob/01cdf40368380d64d90136efb37948c1da79ee0e/src/components/LocationAutocompleteInput/GeocoderMapbox.js#L106)
```
lyyder, line 387: Our code may be subject to change so this link may expire. However, it's good to mention that if the source data only has addresses, then how to turn those into coordinates. Maybe we could mention that coordinate information can be fetched for addresses using a 3rd party geocoding API and link to few options, for example Google and Mapbox, here.

src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
lyyder: For some historical reason I suppose, could be due to the way money is modelled in Go. I think it's fine for now to just take decimals in the intermediary and we can see if it would be helpful to change it at some point.

src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
```diff
@@ -109,12 +215,202 @@ key like this:
...
+reasoning that price in Intermediary is given as decimals but Flex
+handles prices in minor units?_
+
+```
+#:im.listing{...
+            :price #im/money [12.12M "EUR"]
+            ...}
+```
+
+#### UUID
```
lyyder, line 411: It could make sense to communicate that `#uuid` and `#inst` are tagged elements built in to edn, whereas `#im/money` and `#im/location` are intermediary-specific.

src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
```diff
@@ -125,6 +421,12 @@ double information being uploaded to the test site.
 Also, anonymizing test import data by hiding sensitive information like
 names, addresses, email addresses and stripe keys is highly recommended.
 
+We recommend that you only use a subset of your data for the test
+import. The purpose of the test migration is to confirm that your
+extract-and-transform process creates a migration file that is valid and
+consistent, so there is rarely need for importing the anonymized
+equivalent of your full user data into the test environment.
+
```
lyyder, line 452: 👍

src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
```diff
@@ -109,12 +215,202 @@ key like this:
...
+```
+
 ## Things to consider
 
 ### No updates are supported
 
-The import is run only once to the production environment and currently
-updates with multiple imports are not supported.
+The import is loaded only once to the production environment and
+currently updates with multiple imports are not supported.
```
lyyder, line 436: It could be worth mentioning that a migration can be run multiple times to test environment but as updates are not supported, this will result in duplicate data.

lyyder (commented):
Great! Here's a few more findings.

PR #534, opened 2022-02-17
Add listing current stock to attribute list

lyyder:


src/docs/operator-guides/migrating-from-outside-sharetribe/index.md
```diff
@@ -201,6 +202,7 @@ Example syntax:
                       :state :listing.state/pendingApproval
                       :location #im/location [23.12 21.21]
                       :price #im/money [12.12M "EUR"]
+                      :currentStock 5
```
lyyder, line 205: Not sure how much it makes sense to have `currentStock` in these examples as they're not exactly stock listings but it gives an idea how to store stock in intermediary.

lyyder: That could make sense. I'll see if one could be fit in the full example.

lyyder: @SariSaar I'm not able to add a new example listing now. I'll keep this as it is so that it's easy to see how stock is defined even thought the example does not make most sense. Maybe at some point the full example can be updated with better test data and more examples that cover all the edge cases.

PR #536, opened 2022-02-18
Edn lib example

lyyder:
Adds information on different edn libs and some example code on generating intermediary with JS.


lyyder:
The example is now hidden a bit inside `extrainfo`. Maybe it could be also placed in the bottom and linked from where it is now.

SariSaar (approved):
I think an extrainfo box is a good way to implement this!

PR #544, opened 2022-03-07
Add article about listing approval

SariSaar:
Add a background article about the listing approval feature and how it's handled in Flex Console, the FTW templates, and Flex CLI.


shareoc:
Looks really nice @SariSaar! The article is really comprehensive and it's a good idea to include a reference to the Integration API too.

shareoc (approved):

PR #561, opened 2022-05-04
Add Heroku related warnings to relevant articles

SariSaar:
Heroku is having security issues, so add a warning in relevant articles that we're aware of the situation and looking into alternatives.


shareoc (approved):
looks good, lets merge

PR #604, opened 2022-08-17
Add article on reviews

SariSaar:


src/docs/concepts-transactions/reviews/index.md
```diff
@@ -0,0 +1,88 @@
...
+However, you can create a feature that listens to
+[review events](/how-to/reacting-to-events/). Whenever there's a new
+review event, you can then call
+[Flex Integration API](https://www.sharetribe.com/api-reference/integration.html)
+to update the [public metadata](/references/extended-data/#metadata) of
+the reviewed listing.
+
+You can create a logic that stores the number of reviews and their
+previous average in the listing's metadata, and updates these when a new
+review is added. That way, you can e.g. display the average rating on
```
shareoc, line 87:
```suggestion
review is added. That way, you can, e.g. display the average rating on
```

shareoc: just a random thought, but I sometimes wonder if it would be easier for non-native speakers to understand 'for example' instead of e.g.

src/docs/concepts-transactions/reviews/index.md
```diff
@@ -0,0 +1,88 @@
...
+the transaction completing, after which reviews can no longer be added
+through the transaction process. This means that even if one of the
+parties does not submit a review, the other review does get published.
+
+## Frequently asked questions
+
+### Can reviews be removed?
+
+Once published, reviews cannot be removed. They can, however, be
+modified by the operator in Flex Console > Manage > Reviews.
```
shareoc, line 65:
```suggestion
modified by the operator in Flex Console > Manage > [Reviews](https://flex-console.sharetribe.com/reviews).
```

src/docs/concepts-transactions/reviews/index.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Can I add a review after the review period has expired?
+
+Sometimes, transaction participants want to add missing reviews after
+the review period has expired. If the marketplace operators decide that
+the review is relevant and honest, you can reach out to us at
+[Flex Support](mailto:flex-support@sharetribe.com) with the issue and we
+can look into the issue.
```
shareoc, line 73: issue is repeated:
> with the issue and we can look into the issue.

src/docs/concepts-transactions/reviews/index.md
SariSaar: Good point, replaced e.g. with 'for instance' here!

src/docs/concepts-transactions/reviews/index.md
SariSaar: Link added

src/docs/concepts-transactions/reviews/index.md
SariSaar: Good catch, reworded

PR #609, opened 2022-08-19
Start adding discount code article

SariSaar:


shareoc:
I think (with the addition of code examples) this might work better as a how-to or tutorial article. While it's possible to implement discounts using the pricing engine, including them as a Concept article feels like discounts are a built-in feature. Thoughts?

PR #626, opened 2022-10-10
Add a concepts article on integrations in Flex

SariSaar:


src/docs/concepts-integrations/integrations/index.md
```diff
@@ -0,0 +1,268 @@
...
+logic in a secure way.
+
+### Integration API
+
+If you want to create an integration where the third party solution
+listens to, or takes action on, the Flex marketplace data, you will
+likely need to use the Integration API. Where Marketplace API is meant
+for everyday marketplace interactions, Integration API provides
+operators powerful access to all marketplace data and activity. Through
+Integration API, operators can enable third party tools to create and
```
shareoc, line 46: Question: Are you not supposed to use a definite article before "Integration API" and "Marketplace API"?

SariSaar: I guess that's a matter of preference. I think of them as proper names, in which case one wouldn't use an article, but I can double check the usage in our docs and add the articles if they're mostly used with articles.

PR #629, opened 2022-10-21
add info and warning banners

shareoc:
Introduces info and warning banners to Flex Docs. The component can easily be extended if more colours and titles are needed. Can be used by wrapping text inside `<info>` or `<warning>` blocks.

Examples with test content:
Screenshot 2022-10-21 at 11 01 51

Screenshot 2022-10-21 at 11 02 01


SariSaar (approved):
Looks good!

PR #630, opened 2022-10-21
Add mention of Auth0 trailing slash

SariSaar:
Auth0 requires a trailing slash in the IdP URL, but the only way to add that is by reaching out to Flex Support to have it added manually. Add note about the trailing slash and a recommendation to reach out to support.


src/docs/howto-users-and-authentication/enable-open-id-connect-login/index.md
```diff
@@ -18,8 +18,12 @@ another service. This how-to guide assumes that you already have an
 OpenID Connect solution available and intend to use that as a login
 option in your Flex marketplace.
 
-<extrainfo title="Is Apple Sign-in OpenID Connect compliant?">
-Apple Sign-in has several features that resemble the OpenID Connect specification. However, the Apple Sign-in implementation has some differences to the Open ID Connect spec that render it not fully compliant.
+<info>
+
```
shareoc, line 22: @SariSaar FYI For consistent styling, there should be extra whitespace between the <info> tag and where the text begins. The first paragraph, for some reason, isn't rendered in <p> tags if this is omitted.

SariSaar: Gotcha, will keep in mind!

PR #637, opened 2022-11-10
Change extrainfo to info in Stripe non-included countries segment

SariSaar:


SariSaar:
Will need to check the deployment preview URLs (they seem to have /docs/ in the path which breaks in-preview navigation) and images => currently they show up as not loaded in the preview even though they weren't modified in this branch –– is this a feature of the preview or is something else going on?

PR #662, opened 2023-02-16
Add Mac instructions to Flex CLI troubleshooting

SariSaar:


src/docs/introduction/getting-started-with-flex-cli/index.md
```diff
@@ -160,3 +162,9 @@ the PATH environment varible.
 For a step-by-step guide with screenshots, have a look at this blog
 post:
 ['yarn global add' command does not work on Windows](https://sung.codes/blog/2017/12/30/yarn-global-add-command-not-work-windows/)
+
+#### On Mac
+
+1. Open your config file. If you are using _zsh_, it will be `.zshrc`.
```
shareoc, line 168:
```suggestion
1. Open your shell configuration file. If you are using _zsh_, it will be `.zshrc`.
```

src/docs/introduction/getting-started-with-flex-cli/index.md
```diff
@@ -147,11 +147,13 @@ Now that we know how to list processes, the next this is to
...
 ## Troubleshooting
 
-### flex-cli: command not found (on Windows)
+### flex-cli: command not found
 
-If you're seeing `flex-cli: command not found` error on Windows and you
-installed Flex CLI with Yarn, you need to add Yarn global bin path to
-the PATH environment varible.
+If you're seeing `flex-cli: command not found` error and you installed
+Flex CLI with Yarn, you need to add Yarn global bin path to the PATH
```
rap1ds, line 153: Comment: If you want to, you can add link to official Yarn documentation about this: https://classic.yarnpkg.com/en/docs/cli/global/#adding-the-install-location-to-your-path

(Side note! While checking the docs I noticed that the URL is "classic". Apparently, there's a new Yarn version, version 2. It seems that the `global` command isn't there anymore, but instead, the new command is `dlx`. No need to change anything right now but maybe worth checking at some point: https://yarnpkg.com/getting-started/migration#use-yarn-dlx-instead-of-yarn-global) 

(Side note 2. Oh, seems like yarn 2 was released years ago, but still everyone is using version 1?. I have 1.22.10).

rap1ds (approved):

src/docs/introduction/getting-started-with-flex-cli/index.md
SariSaar: Added link!

PR #671, opened 2023-02-28
Improve operator guides

shareoc:


SariSaar:
Looks great! One suggestion: instead of "Concepts" and "Pages", would it make sense to have the latter category be "Content" and also include the one microcopy article that's now in the "Concepts" side. That way, the heading would match what operators see in Flex Console.

As for the other category name, I'm not sure if Concepts is the best one, but I can't think of a better suggestion off the top of my head. Operator resources? General? I'm fine to move forward with Concepts, if we can't think of an alternative.

PR #789, opened 2024-01-31
Clarify deploy-to-render article re: environments

SariSaar:
- Use the term "staging environment" about the deployed application
- Specify when to use client credentials from Dev environment and when to use ones from Test environment


SariSaar:
This PR also contains some formatting changes to the outgoing email settings article, but no content changes have been made there.

PR #814, opened 2024-04-11
Update payout related references

SariSaar:


rap1ds (approved):
stripe-create-payout part: LGTM

bladealslayer (approved):
LGTM. The action reference is usually generated out of the docstrings of the actions in the Core repo, so that should be updated too.

* https://github.com/sharetribe/sharetribe-custom/blob/master/src/sharetribe/tegel/action/privileged_set_line_items.clj#L3
* https://github.com/sharetribe/sharetribe-custom/blob/master/src/sharetribe/tegel/action/stripe_create_payout.clj#L3

PR #839, opened 2024-08-26
Update docs to reflect new access control features

SariSaar:


src/docs/concepts-users-and-authentication/users-and-authentication-in-flex/index.md
```diff
@@ -76,17 +77,26 @@ scheduled to happen automatically.
...
-whether they are registered users or not. If a user wants to start a
-transaction or create a listing, they need to be registered to
-Sharetribe.
+Sharetribe marketplace listings and user profiles on public marketplaces
+can be viewed by anyone by default, whether they are registered users or
+not. If you set your marketplace to private, only registered users can
+view listings and user profiles.
+
+When a user wants to start a transaction or create a listing, they need
+to be registered to Sharetribe. You can also choose to limit additional
```
shareoc, line 86:
```suggestion
to be a registered user on your marketplace. You can also choose to limit additional
```

src/docs/concepts-users-and-authentication/users-and-authentication-in-flex/index.md
```diff
@@ -194,7 +220,10 @@ the email specified in the request.
 
 Only authenticated users can access endpoints that deal with updating
 user information, creating and updating listings, and initiating and
-transitioning transactions.
+transitioning transactions. On private marketplaces, all endpoints
+require an authenticated user, and operators can also further limit
```
shareoc, line 224:
```suggestion
require an authenticated user access token, and operators can also further limit
```

src/docs/concepts-users-and-authentication/user-permissions/index.md
```diff
@@ -14,35 +14,77 @@ permissions to join, post listings, and start transactions. However, you
...
+## Make marketplace private
+
+By default, Sharetribe marketplaces are public. This means that listings
+and user profiles are visible to unauthenticated users. By making your
+marketplace private, you can allow only authenticated users to view
+listings and users on your marketplace.
+
+On a private marketplace, the only public Marketplace API endpoints are
+user creation related [[ TODO CHECK IF THIS IS TRUE ]]. All other
+marketplaces require an authenticated user access token.
```
shareoc, line 28:
```suggestion
endpoints require an authenticated user access token.
```

src/docs/concepts-users-and-authentication/user-permissions/index.md
```diff
@@ -14,35 +14,77 @@ permissions to join, post listings, and start transactions. However, you
...
+
+This setting only has options on the marketplace level.
```
shareoc, line 30: This sentence is a bit ambiguous, I'm not fully sure what this means here 🤔

src/docs/concepts-users-and-authentication/user-permissions/index.md
```diff
@@ -14,35 +14,77 @@ permissions to join, post listings, and start transactions. However, you
...
+
+![Access control for private marketplace](./access_control_private_marketplace.png)
+
+## Approve users who want to join
+
+User approval means that when a user signs in, they need to be approved
```
shareoc, line 36:
```suggestion
User approval means that when a user signs up, they need to be approved
```

src/docs/concepts-users-and-authentication/users-and-authentication-in-flex/index.md
```diff
@@ -28,7 +28,8 @@ you will need to create those limitations in your client application.
 You can use [extended data](/concepts/extended-data-introduction/) to
 determine that a user is in a certain group, and then allow a subset of
 your marketplace functionalities, e.g. listing creation, for a specified
-group.
+group. You can, for example, define user types in Console for this
```
shareoc, line 31: I think linking to help center to the user types article here would be good as we don't have info on user types in docs

src/docs/concepts-users-and-authentication/user-permissions/index.md
```diff
@@ -14,35 +14,77 @@ permissions to join, post listings, and start transactions. However, you
...
 
-![Edit a single user's permissions](./edit-user-permissions.png)
+![Edit a single user's permissions](./edit_user_permissions.png)
 
 Now, you can check or uncheck the permissions checkbox for this user.
 
-![User permissions checkbox](./user-publish-listings-checkbox.png)
+![User permissions checkbox](./user_publish_listings_checkbox.png)
 
 ## Permissions in the currentUser resource
```
shareoc, line 89: Maybe we could mention in this section that you can listen to changes to permissions via Events too

src/docs/concepts-users-and-authentication/user-permissions/index.md
```diff
@@ -14,35 +14,77 @@ permissions to join, post listings, and start transactions. However, you
 can modify some of these permissions in Console, under the "Access
 control" tab.
```
shareoc, line 15: I think here we could have an introductionary paragraph on what this means from a technical perspective => changing these permissions updates the [access-control asset](https://github.com/sharetribe/web-template/commit/e5154d535906ddb4b1d5aabe3a4211be9063e303#diff-2354bc5d624a2c3c0e4a01fa55b9fd8a4e008c0a805a9d7276192ef3f7efce0cR194) and that they are reflected in the currentUser permissions attribute

shareoc: One idea also would explicitly list under each subheading (e.g. Approve users who want to join & Manually grant permission to publish listings) what endpoints this affects (and how) and how the change is reflected in the access-control asset

src/docs/concepts-users-and-authentication/user-permissions/index.md
SariSaar: The other permissions features have a marketplace level toggle and a user level toggle, but this one can only be turned on or off on the marketplace level. Updating the wording here!

src/docs/concepts-users-and-authentication/user-permissions/index.md
SariSaar: Good idea! Added a paragraph on events now.

src/docs/concepts-users-and-authentication/user-permissions/index.md
SariSaar: I added a mention and a screenshot of the access-control.json asset now. However, I didn't add endpoint level descriptions at this point, except to clarify the mention in the private marketplace paragraph. My main reason for this decision was that since this is a concepts article and not a reference, we don't need to include all details on how this feature works. If we do want to include a reference article on access control at some point, I think that would be the correct place for endpoint level details, but I would suggest that we only add that after we have the full feature set out (including any Integration API endpoints for managing permissions) since we will be updating this documentation several times during the coming months.

PR #843, opened 2024-10-02
update user access control article

shareoc:


src/docs/concepts-users-and-authentication/user-permissions/index.md
```diff
@@ -98,12 +110,160 @@ Now, you can check or uncheck the permissions checkbox for this user.
...
+model, you may want to allow only paying users to start transactions. In
+some cases, you may run a marketplace that functions primarily as a
+showroom, where users can browse but cannot buy until the operator
+grants them permission.
+
+You can toggle the setting in the Access control tab.
+
+![Restrict transaction rights checkbox](./users_restrict_transaction_rights.png)
+
+Once this checkbox is selected, similiarly to how publishing rights are
```
SariSaar, line 135:
```suggestion
Once this checkbox is selected, similarly to how publishing rights are
```

src/docs/concepts-users-and-authentication/user-permissions/index.md
```diff
@@ -98,12 +110,160 @@ Now, you can check or uncheck the permissions checkbox for this user.
...
+[Help Center](https://www.sharetribe.com/help/en/articles/9790336).
+
+## API level restrictions
+
+Toggling permissions under the "Access Control" tab in the Console
+directly impacts API behavior. When permissions are modified, endpoint
+access is adjusted accordingly, ensuring that users cannot bypass the
+restrictions imposed by these settings. This guarantees access control
+policies are strictly enforced. These restrictions only apply in the
+Marketplace API.
```
SariSaar, line 152: Would it make sense to add something here to give more context on why it's significant that these restrictions only apply in Marketplace API? Like
- marketplace end-users never access Integration API directly, so Integration API endpoints don't need to be similarly restricted
- if you create an integration where marketplace users indirectly use Integration API, you'll need to enforce the restrictions based on the currentUser effectivePermissionSet similarly to other custom dev
- if you are creating integrations that fetch or update information in Integration API related to third parties, you'd also need to consider how to apply these restrictions in your integration

src/docs/concepts-users-and-authentication/user-permissions/index.md
```diff
@@ -98,12 +110,160 @@ Now, you can check or uncheck the permissions checkbox for this user.
...
+When this setting is toggeld, these endpoints will return a 403
+Forbidden response, indicating that access is denied due to the
+marketplace's private status.
+
+### Approve users who want to join
+
+Toggling this setting in Console will require users to be approved by an
+operator to join the marketplace.
+
+This will restrict the following endpoints:
```
SariSaar, line 179: User approval has slightly different behavior based on whether the marketplace is public or private – this part could refer to the private marketplace endpoints as well somehow, since they're the same for unauth and unapproved users.

src/docs/concepts-users-and-authentication/user-permissions/index.md
```diff
@@ -98,12 +110,160 @@ Now, you can check or uncheck the permissions checkbox for this user.
...
+users can initiate transactions.
+
+When transaction rights have been revoked for a marketplace user, the
+following Marketplace API endpoint will be restricted:
+
+- `POST transactions/initiate`
+
+This endpoint will return a 403 Forbidden response if the user doesn't
+have permission to initiate transactions. Note that users can still
+transact any existing transactions they've initiated.
```
SariSaar, line 238: Maybe a mention here that since transactions/transition is not restricted, it depends on their client app whether users can transact existing transactions – web template blocks e.g. checkout page when coming from TransactionPage in inquiry state, so making a payment on an existing transaction is prevented for non-transaction rights users as well.

SariSaar (approved):
LGTM, a few suggestions and one typo but mainly looks great!

src/docs/concepts-users-and-authentication/user-permissions/index.md
shareoc: @SariSaar great points! Do you have an example of the last bullet point? I'm not entirely sure in what kind of scenario this would be relevant in.

SariSaar: Something like
- you restrict publishing rights behind a paywall
- you want to allow users to post listings on behalf of other users (e.g. org accounts)
- => if someone's subscription has ended and you've revoked their publishing rights, you also want to prevent them from posting on behalf of others

PR #861, opened 2024-11-29
mention that email templates can be disabled

shareoc:


src/docs/concepts-messages/email-notifications/index.md
```diff
@@ -50,6 +50,12 @@ You can preview and customise built-in emails using the
 in the Sharetribe Console. You can find the editor in the Console under
 the Build > Advanced section.
 
+These built-in email notifications can be disabled through Console:
+
+- New message
+- Verify email address
+- User joined
```
rap1ds, line 57: Comment: When we release the new email templates, they should be added here.

thomasmalbaux: Thanks for checking this. Fixed.

PR #873, opened 2025-03-28
Fixed booking slots

shareoc:


src/docs/ftw-availability/availability-management/index.md
```diff
@@ -0,0 +1,138 @@
...
+Sharetribe Web Template, with a focus on technical details and
+implementation. For a general overview of availability management, see
+our
+[Help Center](https://www.sharetribe.com/help/en/articles/8413212-how-availability-management-works).
+See also the
+[reference documentation for availability management](<(/references/availability)>).
+
+## Availability plan types and timezone support
+
+The API supports creating availability plans with either type
```
SariSaar, line 21:
```suggestion
The Sharetribe APIs support creating availability plans with either type
```

src/docs/ftw-availability/availability-management/index.md
```diff
@@ -0,0 +1,138 @@
...
+[query listings](https://www.sharetribe.com/api-reference/marketplace.html#query-listings)
+that use `availabilityPlan/time`, the `availability` parameter needs to
+be set to `time-full` or `time-partial`. Queries using this
+`availability` parameter do not support pagination.
+
+By default, the template will display the first 24 results when using
+the date filter (and as pagination is not supported, the results will be
+limited to 24). This value can be modified in
+[SearchPage.duck.js file](https://github.com/sharetribe/web-template/blob/main/src/containers/SearchPage/SearchPage.duck.js#L21),
+up to a 100.
```
SariSaar, line 45:
```suggestion
so you can allow displaying up to a 100 results.
```

src/docs/ftw-availability/availability-management/index.md
```diff
@@ -0,0 +1,138 @@
...
+
+![Marketplace payment flow](dates-query.png)
+
+## Availability management
+
+### Availability plan
+
+For daily and nightly listings, the template simplifies availability
+management by allowing the provider to select their daily availability.
+Under the hood, availability is automatically set from `00:00` to
```
SariSaar, line 55: Would it make sense to expand the "under the hood", by e.g. linking to the function(s) that do(es) this in the template, or then adding a few sentences on how this happens?

src/docs/ftw-availability/availability-management/index.md
```diff
@@ -0,0 +1,138 @@
...
+creates a listing with the Sharetribe Web Template, the listing is
+always created with type `availabilityPlan/time`, whether the calendar
+booking type is daily, nightly, hourly or fixed.
+
+Using the `availabilityPlan/time` type means that the template can
+specify a certain time zone for the listing's availability plan, which
+is asked when creating a bookable listing. The timezone is associated
+with the availability plan and ensures that all booking times are
+interpreted according to the local time of the listing, rather than the
+time zone of the customer making the booking.
```
Gnito, line 33:
```suggestion
time zone of the customer making the booking - or UTC time zone, which `availabilityPlan/day` uses internally.
```

src/docs/ftw-availability/availability-management/index.md
```diff
@@ -0,0 +1,138 @@
...
+default plan for individual time ranges. Availability exceptions are
+created based on the unit type:
+
+- For daily bookings, exceptions are created per day, and the start and
+  end dates can be the same date
+- For nightly bookings, exceptions are created per night, and the end
+  date needs to be at least one day later than the start date
+- For hourly bookings and fixed-length bookings, exceptions are created
+  with a start time and an end time, and the end time needs to be at
+  least 5 minutes after the start time (both start and end time minutes
```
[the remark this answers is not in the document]
Gnito: To me, this feels that this tries to communicate 2 different things:
- Template: for hourly bookings and fixed-length bookings, start and end times need to match with sharp hours.
- API: in practice, the API endpoint supports shorter exceptions.
  - The smallest exception range is 5 minutes, but the general rule is that start and end times need to map with 5-minute intervals: 00, 05, 10, 15, 20, 25, 30, 35, 40, 45, 50, or 55.

lyyder (commented):
Read the _Interval based filtering_ changes and looks good 👍

Few more comments on top of the ones below:

- There could be a mention that the interval filtering only applies to listings that have an availability plan of type `time`.
- Our time slots resource endpoint is `/timeslots`, as it was one word, but otherwise I think we've referred to them as _time slots_, with a space in between. Not sure if this is consistent across everything though.

src/docs/ftw-availability/availability-management/index.md
```diff
@@ -0,0 +1,138 @@
...
+![Time range](time-range.png)
+
+## Filtering timeslot with intervals
+
+When querying timeslots for a listing using the booking unit "fixed" or
+"hourly", the template uses
+[interval-based filtering](/references/availability). This approach
+allows retrieving significantly less data from the API than using
+regular timeslot queries (the API is restricted to 500 timeslots per
+page). The query used in the template matches the first available
```
Gnito, line 120: Should we give the context: pagination?

```suggestion
regular timeslot queries (the API is restricted to 500 timeslots per paginated
page). The query used in the template matches the first available
```

src/docs/ftw-availability/availability-management/index.md
```diff
@@ -0,0 +1,138 @@
...
+timeslot within the defined interval, and returns that. See the full
+query in
+[ListingPage.duck.js](https://github.com/sharetribe/web-template/blob/main/src/containers/ListingPage/ListingPage.duck.js#L495)
+This information is used to display the availability of a listing in the
+monthly calendar view when booking:
+
+![Time range](april-availability.png)
```
Gnito, line 127: I think the image should contain also dates that are not available!

src/docs/ftw-availability/availability-management/index.md
```diff
@@ -0,0 +1,138 @@
...
+For listings using the fixed booking slot unit type, the template
+extends the end time of the timeslot query. If the duration of the fixed
+booking slot is e.g. 2 hours, the end time of the query is extended by 2
+hours. This allows for the timeslot query to return availability
+correctly for the last moments of the month. For example, to display
+availability on the last day of the month at 23:55 for a 2 hour fixed
+booking slot, the template needs to check if there is availability 2
+hours onwards from 23:55, therefore extending the query into the next
+month. See the
+[relevant code in ListingPage.duck.js](https://github.com/sharetribe/web-template/blob/main/src/containers/ListingPage/ListingPage.duck.js#L488-L493).
```
Gnito, line 138: I think this misses 1 paragraph that states how the date picker works with 'hour' and 'fixed' units. Something like this:

_So, the date picker fetches a reduced set of time slots for the calendar months if the unit type is 'hour' or 'fixed'. This means that, when a user selects one of the days with some availability on the date picker, the app needs to make another query to retrieve all the time slots available on that day._

src/docs/ftw-availability/availability-management/index.md
```diff
@@ -0,0 +1,138 @@
...
+may include listings that are available before or after the time frame
+that the user enters.
+
+![Time range](time-range.png)
+
+## Filtering timeslot with intervals
+
+When querying timeslots for a listing using the booking unit "fixed" or
+"hourly", the template uses
+[interval-based filtering](/references/availability). This approach
```
SariSaar, line 117: Should this point to `/references/availability/#interval-based-filtering` directly, or does it point to the whole article intentionally?

src/docs/ftw-availability/availability-management/index.md
SariSaar: I would even expand the relevance of the pagination here:
- ...this approach allows retrieving significantly less data than regular time slot queries
- a regular timeslot query can fetch up to 500 time slots at once, and to fetch more you'd need to make multiple queries
- with short fixed time slots (e.g. 15 minutes), one day might have dozens of available time slots, so the first 500 time slots of a month might only cover a few weeks
- in the date picker, we're only interested in knowing whether there is availability on a certain day, so we make a query that matches the first available timeslot within the defined interval...

At least for me, it's not intuitive that 500 time slots is not enough for a monthly view of shorter time slots, so an explanation might make sense here

src/docs/references/availability/index.md
```diff
@@ -121,6 +121,87 @@ For time-based plans, both availability exceptions and bookings are
 interpreted literally, i.e. covering the exact time intervals determined
 by their start and end times.
 
+## Interval based filtering
```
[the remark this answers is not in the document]
SariSaar: Maybe starting with a sentence like "timeslots show the listing's actual availability by combining the availability plan with any availability exceptions and bookings". 

Conceptually, I think people sometimes struggle with time slots (customer / anonymous facing) vs availability plans, exceptions, and bookings (author / authenticated facing), so it might make sense to clarify that either here or somewhere else in this article.

src/docs/references/availability/index.md
```diff
@@ -121,6 +121,87 @@ For time-based plans, both availability exceptions and bookings are
...
+Interval based filtering is supported by the
+[/timeslots/query](https://www.sharetribe.com/api-reference/marketplace.html#query-time-slots)
+endpoint. It is controlled via four query parameters:
+
+- `intervalDuration`: the length of each sub-interval (e.g. "PT30M",
+  "P1D")
+- `maxPerInterval`: the maximum number of timeslots to return per
+  sub-interval
+- `minDurationStartingInInterval`: minimum required duration of a
+  timeslot in the interval, in minutes
```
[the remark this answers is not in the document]
SariSaar: Is the difference here that
- min required duration of a time slot **in** the interval → the time slot needs to start after the interval start and end before the interval end
vs
- min required duration of a time slot **after the start of** the interval → the time slot needs to start after the interval start, but the end can be whenever

src/docs/references/availability/index.md
```diff
@@ -121,6 +121,87 @@ For time-based plans, both availability exceptions and bookings are
...
+sub-intervals, which is provided as an
+[ISO 8601 duration](https://www.digi.com/resources/documentation/digidocs/90001488-13/reference/r_iso_8601_duration_format.htm).
+`maxPerInterval` then limits the number of time slots matched within
+each sub-interval. Only a number of results up to the defined maximum
+are returned, with excess slots being excluded.
+
+`minDurationStartingInInterval` sets a threshold (in minutes) that
+determines whether a particular time slot counts toward the
+`maxPerInterval` limit. The threshold value is compared to the duration
+of the overlap between the time slot and the current sub-interval.
```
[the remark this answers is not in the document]
SariSaar: I would amend it to Hannu's suggestion but without the "For example" part, i.e. 
> The threshold value is compared to the duration of the time slot, cut to the start of the current sub-interval. 

This means nothing to people trying to understand this, but it's technically accurate for reference documentation. Example 4 could then use this same phrase "cut to the start of the current sub-interval", to contextualise what this means in practice, e.g.

> Timeslot 1 does not overlap with "Day 2" for 100 minutes or over (it overlaps for 60, as the slot ends at 01:00:00) – the comparison is cut to the start of the current sub-interval. Therefore, the first timeslot that matches the conditions is Timeslot 2, and both timeslots are returned by the query.

SariSaar:
Looks super good 🎉 a lot of work went into these articles! Most of my questions and suggestions here are for adding more context, because this is a really complex topic and I think it'll take people quite a while to wrap their head around some of the concepts. In my opinion, we could err on the side of more explanations instead of conciseness.

src/docs/ftw-availability/availability-management/index.md
shareoc: @Gnito When is UTC used instead of customer time zone?

src/docs/ftw-availability/availability-management/index.md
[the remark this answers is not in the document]
shareoc: @Gnito Double-checking, sharp hours means that value is a factor of an hour, e.g. 15, 20 or 30 minutes?

src/docs/ftw-availability/availability-management/index.md
Gnito: @shareoc to illustrate availabilityPlan/day a bit:

If someone tried to book a listing on New Zealand (2.4.2025 - 4.4.2025) on top of FTW-daily,
what was actually booked was 00:00 moments on UTC, which translates as
2.4.2025 klo 13.00.00 - 4.4.2025 klo 13.00.00 (local time is 13 hours ahead of UTC)

PR #874, opened 2025-04-01
remove mentions of proptypes and defaultprops

shareoc:
The template runs on react 18, no longer any need for proptypes and defaultprops


SariSaar:
LGTM! 👍

PR #877, opened 2025-04-25
remove redundant content and combine some pages

shareoc:


src/docs/ftw-testing-error-handling/how-to-test-ftw/index.md
```diff
@@ -80,7 +82,7 @@ intended to be extended and customized which quickly renders the default
 tests obsolete. The default tests are there to provide a good starting
```
SariSaar, line 82: Now that we also host the web template, we do have a more robust test suite – I'm not sure if we need to mention it here, though. Maybe from the point of view that some pages and features have quite a bit of testing, such as search page filters, and other features don't really have much testing at all. 

I don't have a good suggestion on how to formulate this, so feel free to ignore this comment at your discretion 😄

SariSaar, line 82: Now that we also host the web template, we do have a more robust test suite – I'm not sure if we need to mention it here, though. Maybe from the point of view that some pages and features have quite a bit of testing, such as search page filters, and other features don't really have much testing at all. 

I don't have a good suggestion on how to formulate this, so feel free to ignore this comment at your discretion 😄

PR #884, opened 2025-05-15
Update YouTube embed to use the nocookie version

thomasmalbaux:


thomasmalbaux:
Difficult to test because I can't easily bypass the Cookiebot settings. I'm pretty sure that this is fine though, so I'm making a leap of faith, will check one live, and will revert if needed.

